PR #7, opened 2019-01-15
remove print statements in favor of logging

pipermerriam:
## What was wrong?

Heavy use of `print` statements for logging

## How was it fixed?

Updated to use python `logging` module facilities.

This PR also includes a lot of updates such as:

- whitespace for better clarity
- removing commented out code
- adding `else` clauses to `if/elif/...` blocks which should never reach the final `else`.  New `else` clauses raise exceptions now.
- use of `math.inf` instead of `float("inf")`
- add some `TODO` comments for future cleanup and checks that I don't know how to do now

#### Cute Animal Picture

surprised-animals-monkeys-

### DIFF
--- a/wasm/__init__.py
+++ b/wasm/__init__.py
@@ -622,6 +622,8 @@ def spec_validate_exportdesc(C, exportdesc):
         if len(C["globals"]) <= x:
             raise Exception("invalid")
         mut, t = C["globals"][x]
+        # TODO: verify compliance with the spec for this commented out line.
+        # if mut != "const": raise Exception("invalid") #TODO: this was in the spec, but tests fail linking.wast: $Mg exports a mutable global, seems not to parse in wabt
         return ["global", C["globals"][x]]
     else:
         raise Exception("invalid")
@@ -1793,10 +1795,11 @@ def spec_reinterprett1t2(t1, t2, c):
 
 
 def spec_tconst(config):
-    logger.debug("spec_tconst()")
-
     S = config["S"]
     c = config["instrstar"][config["idx"]][1]
+
+    logger.debug("spec_tconst(%s)", c)
+
     config["operand_stack"] += [c]
     config["idx"] += 1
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
*if* you want to just review the second commit :+1: otherwise we can wait until #6 is merged.

wasm/__init__.py
```diff
@@ -620,13 +619,9 @@ def spec_validate_exportdesc(C, exportdesc):
             raise Exception("invalid")
         return ["mem", C["mems"][x]]
     elif exportdesc[0] == "global":
-        # print("global")
-        # print(len(C["globals"]),x)
         if len(C["globals"]) <= x:
             raise Exception("invalid")
         mut, t = C["globals"][x]
-        # print(mut)
-        # if mut != "const": raise Exception("invalid") #TODO: this was in the spec, but tests fail linking.wast: $Mg exports a mutable global, seems not to parse in wabt
```
carver: Is there an issue tracking this TODO warning? Seems like something that deserves further research.

wasm/__init__.py
```diff
@@ -1807,81 +1793,87 @@ def spec_reinterprett1t2(t1, t2, c):
 
 
 def spec_tconst(config):
-    if verbose >= 1:
-        print("spec_tconst(", ")")
+    logger.debug("spec_tconst()")
```
carver: Looks like it was helpful to include `c` from below in the logs. Maybe move this line down to include the value?

wasm/__init__.py
```diff
@@ -2434,59 +2423,45 @@ def spec_invoke_function_address(config, a=None):
         config["instrstar"] = blockinstrstarendend
         config["idx"] = 0
         config["control_stack"] = []
-        # config_new = {"S":S,"F":F,"instrstar":blockinstrstarendend,"idx":0,"operand_stack":[],"control_stack":[]}
-        # ret = spec_expr(config_new)
-        # if ret=="trap": return ret
```
carver: Yikes at all this commented logic. Maybe the `# TODO confirm correct implementation` is warranted here too.

PR #9, opened 2019-01-16
Use constants for reused powers of 2

pipermerriam:
builds on #7 

## What was wrong?

Inline usage of `2 ** 32` and other similar constants, each of which introduces some runtime overhead.

## How was it fixed?

Converted to use constants

#### Cute Animal Picture

surprised-animals-4


wasm/__init__.py
```diff
@@ -2106,7 +2107,7 @@ def spec_memorygrow(config):
     if sz + n == len(mem["data"]) // 65536:  # success
         config["operand_stack"].append(sz)
     else:
-        config["operand_stack"].append(2 ** 32 - 1)  # put -1 on top of stack
+        config["operand_stack"].append(constants.UINT32_MAX)  # put -1 on top of stack
```
carver: According to the comment, this is really to be interpreted as `INT32_NEGATIVE_ONE`. Using `UINT32_MAX` confused me.

wasm/__init__.py
```diff
@@ -2879,9 +2880,10 @@ def spec_growtable(tableinst, n):
     if verbose >= 1:
         print("spec_growtable(", ")")
     len_ = n + len(tableinst["elem"])
-    if len_ > 2 ** 32:
+
+    if len_ > constants.UINT32_CEIL:
```
carver: Is this off by one? if `len_` needs to fit into a 32 bit uint, then it seems it should be `>=` here.

wasm/__init__.py
```diff
@@ -2895,7 +2897,7 @@ def spec_growmem(meminst, n):
         len(meminst["data"]) % 65536 == 0
     )  # ie divisible by page size = 64 Ki = 65536
     len_ = n + len(meminst["data"]) // 65536
-    if len_ > 2 ** 16:
+    if len_ > constants.UINT16_CEIL:
```
carver: maybe off by one

wasm/__init__.py
```diff
@@ -196,15 +196,15 @@ def spec_validate_functype(ft):
 
 def spec_validate_tabletype(tt):
     limits, elemtype = tt
-    spec_validate_limit(limits, 2 ** 32)
+    spec_validate_limit(limits, constants.UINT32_CEIL)
```
carver: Maybe another off by one, since `spec_validate_limit` uses `>` to compare `k`.

wasm/__init__.py
```diff
@@ -196,15 +196,15 @@ def spec_validate_functype(ft):
...
     return tt
 
 
 # 3.2.4 MEMORY TYPES
 
 
 def spec_validate_memtype(limits):
-    spec_validate_limit(limits, 2 ** 16)
+    spec_validate_limit(limits, constants.UINT16_CEIL)
```
carver: maybe another off by one

wasm/__init__.py
pipermerriam: Added comment to investigate later.  Updating to `UINT32_MAX` resulted in spec test failures.

PR #10, opened 2019-01-16
Piper/use constant for 64k page size

pipermerriam:
Builds on #9

## What was wrong?

Re-use of literal `65536` for 64k page size

## How was it fixed?

Replaced with a constant.

#### Cute Animal Picture

23 shocked surprised

### DIFF
--- a/wasm/constants.py
+++ b/wasm/constants.py
@@ -20,3 +20,7 @@ UINT32_MAX = 2 ** 32 - 1
 INT32_NEGATIVE_ONE = UINT32_MAX
 
 UINT128_CEIL = 2 ** 128
+
+
+# https://webassembly.github.io/spec/core/bikeshed/index.html#memory-instances%E2%91%A0
+PAGE_SIZE_64K = 65536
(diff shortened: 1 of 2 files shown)


wasm/constants.py
```diff
@@ -20,3 +20,6 @@
 INT32_NEGATIVE_ONE = UINT32_MAX
 
 UINT128_CEIL = 2 ** 128
+
+
+PAGE_SIZE_64K = 65536
```
carver: So I think this is: number of pages sized 64,000 bits that fits into a 32-bit uint, right? (or maybe the other way around: there are 64,000 pages that are 65,536 (`2**16`) each? If you wanted to fill it up perfectly, you would do 65,536 pages of 65,536 bits.

Anyway, just trying to figure out where the number comes from, and if we can leave a hint.

pipermerriam: It comes from here: https://webassembly.github.io/spec/core/bikeshed/index.html#memory-instances%E2%91%A0

I updated to include this link in a comment (I hope the docs links don't rot)

PR #18, opened 2019-01-16
Remove magic string based error mechanisms

pipermerriam:
## What was wrong?

Many places in the module use the returning of *magic* strings to signal errors.  This pattern is prone to errors due to it requiring "return value checking" which is very easy to forget.

## How was it fixed?

Finished converting all errors to be exception based.

#### Cute Animal Picture

surprised-animals-12

### DIFF
--- a/wasm/__init__.py
+++ b/wasm/__init__.py
@@ -2760,7 +2760,7 @@ def spec_externtype_matching(externtype1, externtype2):
         if externtype1[1] == externtype2[1]:
             return "<="
         else:
-            raise Unlinkable(f"Extern type mismatch: {externtype1[1]} != {externtype2[1]}")
+            raise Unlinkable(f"Function extern type mismatch: {externtype1[1]} != {externtype2[1]}")
     elif "table" == externtype1[0] and "table" == externtype2[0]:
         limits1 = externtype1[1][0]
         limits2 = externtype2[1][0]
@@ -2770,7 +2770,7 @@ def spec_externtype_matching(externtype1, externtype2):
         if elemtype1 == elemtype2:
             return "<="
         else:
-            raise Unlinkable(f"Element type mismatch: {elemtype1} != {elemtype2}")
+            raise Unlinkable(f"Table element type mismatch: {elemtype1} != {elemtype2}")
     elif "mem" == externtype1[0] and "mem" == externtype2[0]:
         limits1 = externtype1[1]
         limits2 = externtype2[1]
@@ -2784,7 +2784,7 @@ def spec_externtype_matching(externtype1, externtype2):
         if externtype1[1] == externtype2[1]:
             return "<="
         else:
-            raise Unlinkable(f"Extern type mismatch: {externtype1[1]} != {externtype2[1]}")
+            raise Unlinkable(f"Globals extern type mismatch: {externtype1[1]} != {externtype2[1]}")
     else:
         raise Unlinkable(f"Unknown extern type: {externtype1[0]}")
 
@@ -4478,8 +4478,7 @@ def init_store():
 
 
 def decode_module(bytestar):
-    mod = spec_binary_module(bytestar)
-    return mod
+    return spec_binary_module(bytestar)
 
 
 def parse_module(codepointstar):
@@ -4598,7 +4597,7 @@ def type_table(store, tableaddr):
 
 
 def read_table(store, tableaddr, i):
-    if len(store["tables"]) < tableaddr:
+    if len(store["tables"]) <= tableaddr:
         raise ValidationError(
             f"Table address outside of allowed range: {tableaddr} > "
             f"{len(store['tables'])}"
@@ -4750,8 +4749,7 @@ def grow_mem(store, memaddr, n):
 
 
 def alloc_global(store, globaltype, val):
-    store, globaladdr = spec_allocglobal(store, globaltype, val)
-    return store, globaladdr
+    return spec_allocglobal(store, globaltype, val)
 
 
 def type_global(store, globaladdr):
@@ -4786,7 +4784,7 @@ def write_global(store, globaladdr, val):
     # TODO: type check; handle val without type
     gi = store["globals"][globaladdr]
     if gi["mut"] != "var":
-        raise ValidationError("Attempt to write to an immutable global variable")
+        raise ValidationError("Attempt to write to an immutable global variable at address '{globaladdr}'")
     gi["value"] = val
     return store
 
@@ -5353,6 +5351,8 @@ def instantiate_wasm_invoke_start(filename):
 
 
 def instantiate_wasm_invoke_func(filename, funcname, args):
+    # TODO: DRY: this preamble is the same as the
+    # `instantiate_wasm_invoke_start` preamble.
     if not os.path.exists(filename):
         raise ValidationError(f"Unable to open file: {filename}")
 
--- a/wasm/__init__.py
+++ b/wasm/__init__.py
@@ -40,6 +40,8 @@ from wasm.exceptions import (
     InvalidModule,
     MalformedModule,
     Trap,
+    Unlinkable,
+    ValidationError,
 )
 from wasm._utils.types import (
     get_bit_size,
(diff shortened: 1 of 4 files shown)


wasm/__init__.py
```diff
@@ -4602,45 +4599,62 @@ def type_table(store, tableaddr):
 
 def read_table(store, tableaddr, i):
     if len(store["tables"]) < tableaddr:
-        return "error"
+        raise ValidationError(
+            f"Table address outside of allowed range: {tableaddr} > "
+            f"{len(store['tables'])}"
+        )
```
pipermerriam: Lots of repetitive validation code here but I'm inclined to clean this up later rather than trying to DRY it now since I see a pretty clear path to embedding this logic in a handful of `Globals/Tables/Functions/Memories` objects with well defined external APIs that wrap the validation.

wasm/tools/fixtures/generation.py
```diff
@@ -23,7 +23,7 @@ def generate_fixture_tests(metafunc: Any,
     - `fixtures_dir` is the base path that fixture files will be collected from.
     """
     if 'fixture_path' in metafunc.fixturenames:
-        all_fixture_paths = tuple(sorted(find_json_fixture_files(fixtures_dir)))[:58]
+        all_fixture_paths = tuple(sorted(find_json_fixture_files(fixtures_dir)))
```
pipermerriam: Was pleasantly surprised that when I realized this mistake that there was only a single unicode tests which still needed to be cleaned up.

wasm/tools/fixtures/runner.py
```diff
@@ -46,24 +46,26 @@
 logger = logging.getLogger("wasm.tools.fixtures.runner")
 
 
+class FloatingPointNotImplemented(NotImplementedError):
```
pipermerriam: I chose to do it this way to ensure that no other `NotImplementedError` exceptions were caught and interpreted as something that should be ignored.

wasm/__init__.py
```diff
@@ -2742,21 +2744,23 @@ def spec_externtype_matching_limits(limits1, limits2):
...
 
-    if "func" == externtype1[0] and "func" == externtype2[0]:
+    if externtype1[0] != externtype2[0]:
+        raise Unlinkable(f"Mismatch in extern types: {externtype1[0]} != {externtype2[0]}")
+    elif "func" == externtype1[0] and "func" == externtype2[0]:
         if externtype1[1] == externtype2[1]:
             return "<="
         else:
-            raise Exception("Invariant")
+            raise Unlinkable(f"Extern type mismatch: {externtype1[1]} != {externtype2[1]}")
```
carver: Probably want this msg to mention the `func` type

wasm/__init__.py
```diff
@@ -2766,21 +2770,23 @@ def spec_externtype_matching(externtype1, externtype2):
         if elemtype1 == elemtype2:
             return "<="
         else:
-            raise Exception("Invariant")
+            raise Unlinkable(f"Element type mismatch: {elemtype1} != {elemtype2}")
```
carver: mention `table` type

wasm/__init__.py
```diff
@@ -2766,21 +2770,23 @@ def spec_externtype_matching(externtype1, externtype2):
...
         else:
+            # TODO: This code path doesn't appear to be excercised and it
+            # likely isn't an invariant.
             raise Exception("Invariant")
     elif "global" == externtype1[0] and "global" == externtype2[0]:
         if externtype1[1] == externtype2[1]:
             return "<="
         else:
-            raise Exception("Invariant")
+            raise Unlinkable(f"Extern type mismatch: {externtype1[1]} != {externtype2[1]}")
```
carver: mention `global` type

wasm/__init__.py
```diff
@@ -3400,17 +3406,17 @@ def spec_binary_fN_inv(node, N):
...
 # name as UTF-8 codepoints
 def spec_binary_name(raw, idx):
     logger.debug('spec_binary_name()')
     idx, bstar = spec_binary_vec(raw, idx, spec_binary_byte)
 
-    nametxt = bytearray(bstar).decode()
+    try:
+        nametxt = bytearray(bstar).decode()
+    except UnicodeDecodeError as err:
+        raise MalformedModule from err
```
carver: :+1:

wasm/__init__.py
```diff
@@ -4472,14 +4478,7 @@ def init_store():
...
 def decode_module(bytestar):
-    try:
-        mod = spec_binary_module(bytestar)
-    except AssertionError:
-        raise
-    except EXCEPTIONS_TO_RERAISE:
-        raise
-    except Exception as err:
-        return "malformed"
+    mod = spec_binary_module(bytestar)
```
carver: Sweet, maybe return in the same line.

Looks like `decode_module` is an alias for `spec_binary_module` now, which might have other knock-on effects (not for this PR).

wasm/__init__.py
```diff
@@ -4602,45 +4599,62 @@ def type_table(store, tableaddr):
 
 def read_table(store, tableaddr, i):
     if len(store["tables"]) < tableaddr:
-        return "error"
+        raise ValidationError(
```
carver: Huh, the other tests like `type_table` were testing `<=`. Not sure which one is right, but they should almost certainly be the same.

I guess that will get taken care of in the the follow-up refactor you mentioned. But maybe a `#TODO` note until then.

wasm/__init__.py
```diff
@@ -4602,45 +4599,62 @@ def type_table(store, tableaddr):
...
     if type(i) != int or i < 0:
-        return "error"
+        raise ValidationError(
+            f"Invalid table index.  Must be positive integer.  Got {repr(i)}"
+        )
     ti = store["tables"][tableaddr]
     if i >= len(ti["elem"]):
-        return "error"
+        raise ValidationError(
+            f"Index out of range for table.  {i} >= {len(ti['elem'])}"
```
carver: To match the pattern from the other exceptions, looks like this should say `{i} < {len...`.

wasm/__init__.py
```diff
@@ -4716,18 +4750,16 @@ def grow_mem(store, memaddr, n):
 
 
 def alloc_global(store, globaltype, val):
-    try:
-        store, globaladdr = spec_allocglobal(store, globaltype, val)
-    except EXCEPTIONS_TO_RERAISE:
-        raise
-    except:
-        return store, "error"
+    store, globaladdr = spec_allocglobal(store, globaltype, val)
```
carver: Another one that can be a one-line return/alias.

wasm/__init__.py
```diff
@@ -4736,19 +4768,25 @@ def type_global(store, globaladdr):
...
-        return "error"
+        raise ValidationError(
+            f"Globals address outside of allowed range: {globaladdr} > "
+            f"{len(store['globals'])}"
+        )
     # TODO: type check; handle val without type
     gi = store["globals"][globaladdr]
     if gi["mut"] != "var":
-        return "error"
+        raise ValidationError("Attempt to write to an immutable global variable")
```
carver: Maybe include the `globaladdr` in the msg?

wasm/__init__.py
```diff
@@ -5296,41 +5334,46 @@ def iterate_through_expression_and_validate_each_opcode(
...
-        return "error, could not read " + filename
-    module = decode_module(bytestar)  # get module as abstract syntax
+    if not os.path.exists(filename):
+        raise ValidationError(f"Unable to open file: {filename}")
+
+    with open(filename, 'rb') as file_:
+        bytestar = memoryview(file_.read())
+        if not bytestar:
+            raise ValidationError(f"Error reading file: {filename}")
+        module = decode_module(bytestar)  # get module as abstract syntax
```
carver: Looks like a solid DRY opportunity here, `instantiate_wasm_invoke_start` and this both have long, identical (I think) preambles. Maybe for a `#TODO`

wasm/tools/fixtures/runner.py
```diff
@@ -72,19 +74,16 @@ def instantiate_module_from_wasm_file(
...
+
             for export in sub_module["exports"]:
                 if export["name"] == import_["name"]:
                     externval = export["value"]
-            if externval is None:
+                    break
+            else:
                 raise Unlinkable("Unlinkable module: export name not found")
-            if externval[0] != import_["desc"][0]:
-                raise Unlinkable("Unlinkable module: TODO figure out what this means")
```
carver: Deleting `TODO figure out what this means` when I don't know what it means scares me.

pipermerriam: Basically this code path wasn't being used because this validation now happens internally with the `Unlinkable` exception originating from inside the module, and I'm reasonably confident that it now has a friendly/useful error message.

PR #19, opened 2019-01-17
Re organize module code

pipermerriam:
Builds on #18 

## What was wrong?

All of the main module code was located in `wasm.__init__` which made it a bit difficult to enforce type hinting and linting.

## How was it fixed?

Moved logic to live in `wasm.main` (which is intended to be temporary).

This also:

- relocates the `LICENSE` form being inline to be in the `LICENSE` file.

#### Cute Animal Picture

surprised-animals-8


carver:
:+1:

PR #20, opened 2019-01-17
Fill in type hinting for non core library logic and add mypy to CI

pipermerriam:
Builds on #19 

## What was wrong?

No `mypy` in ci and missing type hints

## How was it fixed?

Add type hints to everything but `wasm.main` as well as CI run to run type checks on all modules except `wasm.main`

#### Cute Animal Picture

surprised-animals-15

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,9 +6,6 @@ from setuptools import (
 )
 
 extras_require = {
-    'wasm': [
-        "mypy_extensions>=0.4.1,<1.0.0",
-    ],
     'test': [
         "pytest==4.1.0",
         "pytest-watch==4.2.0",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -47,9 +51,7 @@
     author_email='ethcalibur+pip@gmail.com',
     url='https://github.com/ethereum/py-wasm',
     include_package_data=True,
-    install_requires=[
-        "eth-utils>=1,<2",
-    ],
+    install_requires=extras_require['wasm'],
```
carver: I don't follow the concept here. Is the idea to use the `wasm` extra as the same thing as the core requirements for the module?

It's mostly weird because it will create another extra, like `wasm[wasm]` that is a noop. Not exactly a problem, just odd.

Seems like `install_requires=['mypy_extensions...` isn't a problem, and matches our other standard practices better.

pipermerriam: Following a pattern we're using elsewhere:

- https://github.com/ethereum/py-evm/blob/6abf213883970a94fcd2427a106c1a6e80bf4970/setup.py#L83
- https://github.com/ethereum/trinity/blob/master/setup.py#L92

But I tend to agree, going to revert to inlining as suggested.

carver: Oooh, my understanding was that we were originally doing that because of the "two projects in one repo" scenario, like `py-evm[trinity]`. Maybe that unintentionally survived the repo split.

PR #22, opened 2019-01-17
Piper/implement formal data structure for table type

pipermerriam:
Builds on #21 

## What was wrong?

The Web Assembly spec defines a data structure *Table Type*: https://webassembly.github.io/spec/core/bikeshed/index.html#table-types%E2%91%A0

This was being internally represented as a length-2 list:  `[Limits(...), "anyfunc"]`

This data structure is mutable **and** it uses a magic string to represent the element type.

## How was it fixed?

Added a *stub* data type `FuncType` to be used as the element type and then updated all representations of a table type to use a new `TableType` data structure which is a `NamedTuple`.

#### Cute Animal Picture

surprised-shocked-animals-funny-3__700

### DIFF
--- a/wasm/main.py
+++ b/wasm/main.py
@@ -17,7 +17,7 @@ from wasm._utils.types import (
     is_integer_type,
 )
 from wasm.datatypes import (
-    FuncType,
+    FuncRef,
     Limits,
     TableType,
 )
@@ -443,7 +443,7 @@ def spec_validate_call_indirect(C, x):
     if C["tables"] == None or len(C["tables"]) < 1:
         raise InvalidModule("invalid")
     limits, elemtype = C["tables"][0]
-    if elemtype != FuncType:
+    if elemtype is not FuncRef:
         raise InvalidModule("invalid")
     if C["types"] == None or len(C["types"]) <= x:
         raise InvalidModule("invalid")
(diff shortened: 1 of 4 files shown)


wasm/main.py
```diff
@@ -438,7 +443,7 @@ def spec_validate_call_indirect(C, x):
     if C["tables"] == None or len(C["tables"]) < 1:
         raise InvalidModule("invalid")
     limits, elemtype = C["tables"][0]
-    if elemtype != "anyfunc":
+    if elemtype != FuncType:
```
davesque: Should this be `elemtype is not FuncType`?

davesque: Actually, here's another thing: down in the "Binary Format" section of the w3c spec, it appears that types are actually encoded into modules as metadata for validity checking.  According to that section of the spec, the `funcref` elemtype has a specific encoding of `0x70` (https://webassembly.github.io/spec/core/bikeshed/index.html#table-types%E2%91%A4).  Therefore, it actually seems appropriate to me that `elemtype` should refer to a singleton value called something like `FuncRef` and not `FuncType`.  In the same section, function types have their own separate encodings based on the vectors of input and output types (https://webassembly.github.io/spec/core/bikeshed/index.html#function-types%E2%91%A4).

davesque: Another way to put it is that the concepts of `funcref` and the type construct of "function types", which might be referred to by a label such as `FuncType`, seem to be regarded differently in terms of the spec even though the spec says that `funcref` is meant to represent the union of function types.  The `funcref` label still seems to be used fairly often and corresponds more or less with a particular piece of encoded metadata.

pipermerriam: I'm convinced.

PR #27, opened 2019-01-18
Piper/implement formal data structure for exports type

pipermerriam:
## What was wrong?

The WebAssembly spec defines a formal data type for module exports.  Currently this is modeled with a combination of dictionary and lists, both mutable data structures, and both somewhat awkard to deal with.

## How was it fixed?

The Export type was implemented using a `NamedTuple`.  The `desc` field is a typed integer index/address for any of `funcidx/tableidx/memidx/globalidx`.  This was accomplished using four subclasses of the built-in `int` type.

#### Cute Animal Picture

surprised-animals-12


wasm/main.py
```diff
@@ -2901,35 +2958,35 @@ def spec_allocmodule(S, module, externvalimstar, valstar):
...
+            desc = funcaddrmodstar[exporti.desc]
+        elif exporti.is_table:
+            desc = tableaddrmodstar[exporti.desc]
+        elif exporti.is_memory:
+            desc = memaddrmodstar[exporti.desc]
+        elif exporti.is_global:
+            desc = globaladdrmodstar[exporti.desc]
         else:
-            raise Exception("Invariant: TODO: bettermessage")
+            raise Exception("Invariant: Unknown export ")
```
pipermerriam: Exception message should contain the type of `exporti`

wasm/main.py
```diff
@@ -609,31 +639,34 @@ def spec_validate_start(C, start):
...
-    if exportdesc[0] == "func":
-        if len(C["funcs"]) <= x:
+def spec_validate_export(C: Context, export: Export) -> TExportValue:
+    return spec_validate_exportdesc(C, export.desc)
+
+
+def spec_validate_exportdesc(C: Context,
+                             desc: ExportDesc) -> TExportValue:
+    if isinstance(desc, FuncIdx):
+        if len(C["funcs"]) <= desc:
```
carver: I don't follow what `desc` is supposed to mean here. How about something like `export_index`? (Same for the type name)

wasm/main.py
```diff
@@ -609,31 +639,34 @@ def spec_validate_start(C, start):
...
             raise InvalidModule("invalid")
-        return ["func", C["funcs"][x]]
-    elif exportdesc[0] == "table":
-        if len(C["tables"]) <= x:
+        return C["funcs"][desc]
```
carver: Huh, interesting that the caller didn't explicitly need to have this type hint handed to them `"func"`. I guess the caller always knows what kind of value it is looking up?

wasm/main.py
```diff
@@ -609,31 +639,34 @@ def spec_validate_start(C, start):
...
-        return ["mem", C["mems"][x]]
-    elif exportdesc[0] == "global":
-        if len(C["globals"]) <= x:
+        return C["mems"][desc]
+    elif isinstance(desc, GlobalIdx):
+        if len(C["globals"]) <= desc:
             raise InvalidModule("invalid")
-        mut, t = C["globals"][x]
         # TODO: verify compliance with the spec for this commented out line.
+        #mut, t = C["globals"][desc]
```
carver: :confused: intentional?

wasm/main.py
```diff
@@ -720,70 +753,86 @@ def spec_validate_module(mod):
...
-            if len(C["funcs"]) <= export["desc"][1]:
+        if export.is_function:
+            if len(C["funcs"]) <= export.desc:
                 # this was not explicit in spec
                 raise InvalidModule("invalid")
-            etstar.append(["func", C["funcs"][export["desc"][1]]])
-        elif export["desc"][0] == "table":
-            if len(C["tables"]) <= export["desc"][1]:
+            etstar.append(C["funcs"][export.desc])
+        elif export.is_table:
```
carver: Nice for this whole area :+1:

wasm/tools/fixtures/modules.py
```diff
@@ -73,23 +78,31 @@ def spectest__print(store, arg):
             FuncType((ValType.f64, ValType.f64), ()),
             FuncType((), ()),
         ],
-        "funcaddrs": [0, 1, 2, 3, 4, 5, 6],
-        "tableaddrs": [0],
-        "memaddrs": [0],
-        "globaladdrs": [0, 1, 2],
+        "funcaddrs": [
```
carver: Maybe `list(map(FuncIdx, range(7)))`?

Maybe that's too much, instead: `list(map(FuncIdx, [0, 1, 2, 3, 4, 5, 6]))`

wasm/main.py
pipermerriam: I went with `index` which I'm very meh about but I see this as something that we can reasonably easily evolve the name as we find better ones.  Index for Indices which is what the value is, an index into the declared functions/tables/memory/globals.

wasm/tools/fixtures/modules.py
pipermerriam: Went with a list comprehension.  I feel icky ignoring all this mutable *everything*

wasm/tools/fixtures/modules.py
```diff
@@ -73,23 +78,23 @@ def spectest__print(store, arg):
             FuncType((ValType.f64, ValType.f64), ()),
             FuncType((), ()),
         ],
-        "funcaddrs": [0, 1, 2, 3, 4, 5, 6],
-        "tableaddrs": [0],
-        "memaddrs": [0],
-        "globaladdrs": [0, 1, 2],
```
carver: Looks like `GlobalIdx(2)` got dropped. Intentional?

PR #28, opened 2019-01-19
Implement formal data structure for imports.

pipermerriam:
## What was wrong?

The WebAssembly spec defines a formal data structure for Imports.  Current codebase used a mixture of lists and dictionaries to represent them which results in heavy mutability and difficult readability.

## How was it fixed?

Implemented an Import type in roughly the same manner as #27 with a `NamedTuple` for the Import data structure and an additional `int` subclass, `TypeIdx` for the new indices type that needed to be supported.

#### Cute Animal Picture

14628390535_503e98f8e1_b

### DIFF
--- a/wasm/main.py
+++ b/wasm/main.py
@@ -714,7 +714,7 @@ def spec_validate_module(mod: Module) -> List[List[ExternType]]:
     itstar: List[ExternType] = []
     for import_ in mod["imports"]:
         if import_.is_function:
-            if len(mod["types"]) <= import_.desc:
+            if import_.desc >= len(mod["types"]):
                 # this was not explicit in spec
                 raise InvalidModule(
                     f"Function import out of range: {import_.desc} > "
@@ -2805,29 +2805,29 @@ def spec_externtype_matching(externtype1, externtype2):
         raise Unlinkable(
             f"Mismatch in extern types: {type(externtype1)} != {type(externtype2)}"
         )
-    elif isinstance(externtype1, FuncType) and isinstance(externtype2, FuncType):
+    elif isinstance(externtype1, FuncType):
         if externtype1 == externtype2:
             return "<="
         else:
             raise Unlinkable(f"Function types not equal: {externtype1} != {externtype2}")
-    elif isinstance(externtype1, TableType) and isinstance(externtype2, TableType):
+    elif isinstance(externtype1, TableType):
         spec_externtype_matching_limits(externtype1.limits, externtype2.limits)
 
-        if externtype1.elem_type is externtype1.elem_type:
+        if externtype1.elem_type is externtype2.elem_type:
             return "<="
         else:
             raise Unlinkable(
                 f"Table element type mismatch: {externtype1.elem_type} != "
                 f"{externtype2.elem_type}"
             )
-    elif isinstance(externtype1, MemoryType) and isinstance(externtype2, MemoryType):
+    elif isinstance(externtype1, MemoryType):
         if spec_externtype_matching_limits(externtype1, externtype2) == "<=":
             return "<="
         else:
             # TODO: This code path doesn't appear to be excercised and it
             # likely isn't an invariant.
             raise Exception("Invariant")
-    elif isinstance(externtype1, GlobalType) and isinstance(externtype2, GlobalType):
+    elif isinstance(externtype1, GlobalType):
         if externtype1 == externtype2:
             return "<="
         else:
@@ -4565,8 +4565,9 @@ def module_imports(module):
         )
 
     result = []
-    for importi, importstar in zip(importstar, externtypestar):
-        resutli = [immporti.module, importi.name, externtypei]
+    for importi, externtypei in zip(importstar, externtypestar):
+        # This code path appears to be missing test coverage.
+        resutli = [importi.module, importi.name, externtypei]
         result += resulti
     return result
 
--- a/wasm/main.py
+++ b/wasm/main.py
@@ -25,12 +25,14 @@ from wasm.datatypes import (
     FuncType,
     GlobalIdx,
     GlobalType,
+    Import,
     Limits,
     MemoryIdx,
     MemoryType,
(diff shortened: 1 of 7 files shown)


wasm/main.py
```diff
@@ -676,61 +664,72 @@ def spec_validate_exportdesc(C: Context,
...
+    ttstar: List[TableType] = [table["type"] for table in mod["tables"]]
+    mtstar: List[MemoryType] = [mem["type"] for mem in mod["mems"]]
+    gtstar: List[GlobalType] = [global_["type"] for global_ in mod["globals"]]
+
+    itstar: List[ExternType] = []
     for import_ in mod["imports"]:
-        if import_["desc"][0] == "func":
-            if len(mod["types"]) <= import_["desc"][1]:
+        if import_.is_function:
+            if len(mod["types"]) <= import_.desc:
```
carver: *nit* reads like a yoda conditional to me

wasm/main.py
```diff
@@ -2811,40 +2801,39 @@ def spec_externtype_matching_limits(limits_a: Limits, limits_b: Limits) -> str:
 def spec_externtype_matching(externtype1, externtype2):
     logger.debug('spec_externtype_matching(%s, %s)', externtype1, externtype2)
 
-    if externtype1[0] != externtype2[0]:
-        raise Unlinkable(f"Mismatch in extern types: {externtype1[0]} != {externtype2[0]}")
-    elif "func" == externtype1[0] and "func" == externtype2[0]:
-        if externtype1[1] == externtype2[1]:
+    if type(externtype1) is not type(externtype2):
```
carver: I can't think of a reason this would be a problem, but `is` tests usually stand out to me as needing a reason. I can imagine a python implementation (like PyPy) breaking in weird ways. So `==` seems life the safer, standard bet, but it's really just intuition.

wasm/main.py
```diff
@@ -2811,40 +2801,39 @@ def spec_externtype_matching_limits(limits_a: Limits, limits_b: Limits) -> str:
...
+        raise Unlinkable(
+            f"Mismatch in extern types: {type(externtype1)} != {type(externtype2)}"
+        )
+    elif isinstance(externtype1, FuncType) and isinstance(externtype2, FuncType):
```
carver: Hm, checking them both after showing that they are equal feels redundant. No objection really, but I'd be happy with skipping the second check, too.

wasm/main.py
```diff
@@ -2811,40 +2801,39 @@ def spec_externtype_matching_limits(limits_a: Limits, limits_b: Limits) -> str:
...
+            raise Unlinkable(f"Function types not equal: {externtype1} != {externtype2}")
+    elif isinstance(externtype1, TableType) and isinstance(externtype2, TableType):
+        spec_externtype_matching_limits(externtype1.limits, externtype2.limits)
 
-        limits_a, limits_b = table_type_a.limits, table_type_b.limits
-        spec_externtype_matching_limits(limits_a, limits_b)
-
-        elem_type_a, elem_type_b = table_type_a.elem_type, table_type_b.elem_type
-        if elem_type_a is elem_type_b:
+        if externtype1.elem_type is externtype1.elem_type:
```
carver: It seems like this one should be an `==`, since another instance of the `NamedTuple` should match if all the values match (I would think).

Also, I assume the second variable name is a typo.
```suggestion
        if externtype1.elem_type == externtype2.elem_type:
```

wasm/main.py
```diff
@@ -4581,11 +4563,10 @@ def module_imports(module):
             f"Wrong import length: expected {len(extertypeprimestar)} / got "
             f"{len(externtypestar)}"
         )
+
     result = []
-    for i in range(len(importstar)):
-        importi = importstar[i]
-        externtypei = externtypestar[i]
-        resutli = [immporti["module"], importi["name"], externtypei]
+    for importi, importstar in zip(importstar, externtypestar):
```
carver:
```suggestion
    for importi, externtypei in zip(importstar, externtypestar):
```

wasm/main.py
```diff
@@ -4581,11 +4563,10 @@ def module_imports(module):
...
+        resutli = [immporti.module, importi.name, externtypei]
```
carver:
```suggestion
        resutli = [importi.module, importi.name, externtypei]
```
Looks like this method should be commented as missing test coverage, too.

wasm/main.py
```diff
@@ -4599,11 +4580,12 @@ def module_exports(module):
     if len(exportstar) != len(externtypeprimestar):
         raise Exception("TODO: proper error message")
 
+    # TODO: this code path may not be excercised. Verify.
```
carver: :+1:

wasm/main.py
pipermerriam: same reason why we use `if value is None`.  Since classes are singletons I have historically used the convention of using `is` checks  for equality between classes.

wasm/main.py
pipermerriam: `elem_type` is a class, not an instance so defaulted to `is` check (see other comment for why).

wasm/main.py
davesque: Not that it matters really but...
```suggestion
    # TODO: this code path may not be exercised. Verify.
```

wasm/main.py
carver: > classes are singletons

Ah, didn't realize this was a python runtime guarantee. I couldn't figure out how to google for it. The closest I could find was this [SO post](https://stackoverflow.com/a/51503803/8412986):
>  class equality tests only ever return true for objects that are identical. reference_to_class == other_reference_to_class is only ever True if both references are to the same class object; the test is essentially the same as reference_to_class is other_reference_to_class.

Good enough for me :+1:

wasm/main.py
carver: :+1: yeah, somehow I thought it was an instance :man_facepalming:

PR #29, opened 2019-01-22
Implement formal data structures for Instruction types.

pipermerriam:
### What was wrong

Parsing of raw wasm bytecode to an internal representation suitable for validation and execution was previous done with a large function body with many nested `if/elif/else` statements, resulting in instructions in the form of `["<opcode-text>", ...<args>]` where `<args>` were things like the parsed constant values for opcodes like `i32.const` or the list of nested instructions in the case of opcodes like `block`.

Subsequently, to do things like determining the argument types of an instructions, things like `typ = 1instr[0][:3]` would occur to infer the type from the text representation of the opcode.

### How was this fixed.

The parsing of the opcodes is now done with a single `parse_instruction` function which delegates to multiple sub-parsers.

The instructions themselves are now represented using formal classes which also implement strongly typed values for things like value types, whether the instruction is signed, etc.  This allows for `isinstance` checks or direct comparison against the `instruction.opcode` and the `Enum` of all opcodes.

funny-pictures-surprised-animals-08

### DIFF
--- a/wasm/datatypes/function.py
+++ b/wasm/datatypes/function.py
@@ -1,0 +1,26 @@
+from typing import (
+    TYPE_CHECKING,
+    NamedTuple,
+    Tuple,
+)
+
+from .indices import (
+    TypeIdx,
+)
+from .val_type import (
+    ValType,
+)
+
+if TYPE_CHECKING:
+    from wasm.instructions import (  # noqa: F401
+        Instruction,
+    )
+
+
+class ModuleFunction(NamedTuple):
+    """
+    https://webassembly.github.io/spec/core/bikeshed/index.html#functions%E2%91%A0
+    """
+    type: TypeIdx
+    locals: Tuple[ValType, ...]
+    expr: Tuple['Instruction', ...]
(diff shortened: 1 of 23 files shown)


wasm/datatypes/function.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+
+class ModuleFunction(NamedTuple):
+    """
+    https://webassembly.github.io/spec/core/bikeshed/index.html#functions%E2%91%A0
+    """
+    type: TypeIdx
+    locals: Tuple[ValType, ...]
+    expr: Tuple[Instruction, ...]
```
pipermerriam: use `wasm.typing.Expression`

wasm/main.py
```diff
@@ -784,7 +834,7 @@ def spec_validate_module(mod: Module) -> List[List[ExternType]]:
     for i, func in enumerate(mod["funcs"]):
         ft = spec_validate_func(C, func)
         # TODO: remove list cast once everything is a tuple
```
pipermerriam: update comment.

pipermerriam:
@davesque ready for review.

pipermerriam:
I believe there is a `ModuleFunction` type in this PR that I can extract as it is really part of a subsequent WIP in #42

PR #30, opened 2019-01-25
Opcode data structures

pipermerriam:
Extracted from https://github.com/ethereum/py-wasm/pull/29

## What was wrong?

Need a standardized way to reference opcodes, both in their binary and text representations.

## How was it fixed?

This PR simply creates two data structures.

* An `Enum` that holds the binary representations.
* A mapping from the `Enum` values to their text representations.

The `BinaryOpcode` data structure also implements multiple convenience properties for determining what category an opcode falls in.

#### Cute Animal Picture

16 -surprised


pipermerriam:
@carver nice part about current structure is that you can do `opcode = BinaryOpcode(0x00)` which is a very useful API.

PR #31, opened 2019-01-25
Add singleton class

pipermerriam:
Extracted from #29 

## What was wrong?

Many of the opcodes are stateless meaning that we don't actually need to instantiate a new object for each occurrence of the opcode instruction.

## How was it fixed?

This implements a simple `Singleton` base class which keeps a reference to the first instance created for a given class and then always returns the same instance on subsequent instantiations.

#### Cute Animal Picture

animals-also-get-surprised-028

### DIFF
--- a/tests/core/test_singleton_util.py
+++ b/tests/core/test_singleton_util.py
@@ -1,0 +1,51 @@
+from wasm._utils.singleton import (
+    Singleton,
+)
+
+
+def test_base_singleton_class():
+    inst_a = Singleton()
+    inst_b = Singleton()
+
+    assert inst_a is inst_b
+
+
+class SubA(Singleton):
+    pass
+
+
+class SubB(Singleton):
+    pass
+
+
+def test_subclassed_singletons():
+    base = Singleton()
+
+    inst_a1 = SubA()
+    inst_a2 = SubA()
+
+    inst_b1 = SubB()
+    inst_b2 = SubB()
+
+    assert inst_a1 is inst_a2
+    assert inst_b1 is inst_b2
+
+    assert base is not inst_a1
+    assert base is not inst_b1
+    assert inst_a1 is not inst_b1
+
+
+class WithArgs(Singleton):
+    def __init__(self, val_a, val_b):
+        self.val_a = val_a
+        self.val_b = val_b
+
+
+def test_singleton_class_with_initialization_args():
+    inst_a = WithArgs(1, val_b='2')
+    inst_b = WithArgs(1, val_b='2')
+
+    assert inst_a is inst_b
+
+    assert inst_a.val_a == 1
+    assert inst_a.val_b == '2'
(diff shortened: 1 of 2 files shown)


pipermerriam:
I don't believe this approach suffers from any of those downsides but I'm open to alternate patterns to accomplish the same thing (reduce the number of python objects we have to instantiate).

Your suggestion about a registry doesn't work because:

1) the `i32.const`, `i64.const`, `f32.const`, and `f64.const` each have `~2**N` possible  values that can be instantiated with `N` being `32` or `64`.  We **could** cache these in hopes to re-use the same instance for common values.
2) the `block`, and `if` instructions actually wrap another instruction expression and must be individually instantiated.
3) the instructions for loading and retrieving values from memory have values associated with them and must each be instantiated.  Same caching story as the `i32.const` instructions.

But, that's probably only 10-20% of the opcodes.   The other 80% could all be instantiated once and reused.  I'm going to go forward with this pattern as is and will copy this info to a new issue ticket for future improvements.

tests/core/test_singleton_util.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_singleton_class_with_initialization_args():
+    inst_a = WithArgs(1, val_b='2')
+    inst_b = WithArgs(1, val_b='2')
+
+    assert inst_a is inst_b
+
+    assert inst_a.val_a == 1
+    assert inst_a.val_b == '2'
```
lithp: This might also include a test that `inst_b = WithArgs(2, val_b='4')` is still `inst_a`, and also that `(inst_a.val_a, inst_a.val_b)` has become `(2, 4)` .

pipermerriam: Yeah, this demonstrates that the current singleton implementation is broken for something that has args.  Will address separately, probably just by enforcing that it cannot have any args/kwargs

wasm/_utils/singleton.py
```diff
@@ -0,0 +1,16 @@
...
+)
+
+
+class Singleton:
+    _cache: Dict[Type['Singleton'], 'Singleton'] = {}
+
+    def __new__(cls, *args: Any, **kwargs: Any) -> 'Singleton':
+        if cls not in cls._cache:
+            instance = super().__new__(cls)
+            instance.__init__(*args, **kwargs)
```
lithp: I think there's a foot-gun here in that the result of `__new__()` [will have `__init__()` called upon it](https://docs.python.org/3/reference/datamodel.html#object.__new__). For the first instance of the singleton `__init__()` is called twice! Subsequent calls to the Singleton will return the same object, but then `__init__()` will be called on that object. Code which assumes that "single instance" means "single initialization" will be buggy, and I think that's a reasonable assumption to make.

pipermerriam: Fixed in https://github.com/ethereum/py-wasm/pull/38

PR #32, opened 2019-01-25
Add bitsize enum

pipermerriam:
Extracted from #29 

## What was wrong?

Using an integer for bit size was less precise than desired.

## How was it fixed?

Updated to be an `Enum` for better constraints on the allowed values.

#### Cute Animal Picture

surprised-animals-16

### DIFF
--- a/wasm/datatypes/indices.py
+++ b/wasm/datatypes/indices.py
@@ -6,6 +6,14 @@ class GlobalIdx(int):
     pass
 
 
+class LabelIdx(int):
+    pass
+
+
+class LocalIdx(int):
+    pass
+
+
 class MemoryIdx(int):
     pass
 
--- a/wasm/main.py
+++ b/wasm/main.py
@@ -19,6 +19,7 @@ from wasm._utils.validation import (
     get_duplicates,
 )
 from wasm.datatypes import (
+    BitSize,
     Export,
     FuncIdx,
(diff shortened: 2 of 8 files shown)


carver:
Is this a writeup for a different branch?

pipermerriam:
@carver updated

wasm/datatypes/indices.py
```diff
@@ -6,6 +6,14 @@ class GlobalIdx(int):
...
 
 
+class LabelIdx(int):
+    pass
+
+
+class LocalIdx(int):
+    pass
+
+
```
pipermerriam: Tossed these in as they are innocuous and it gets them out of the big PR.

wasm/main.py
```diff
@@ -3452,7 +3452,7 @@ def spec_binary_fN(raw, idx, N):
     for i in range(N // 8):
         bstar += bytearray([raw[idx]])
         idx += 1
-    return idx, spec_bytest_inv(ValType.get_float_type(N), bstar)  # bytearray(bstar)
+    return idx, spec_bytest_inv(ValType.get_float_type(BitSize(N)), bstar)  # bytearray(bstar)
```
carver: Yay for automatically forcing `N` to be one of the values, as part of the `enum` initialization :+1:

PR #39, opened 2019-01-26
Implement Interned base class and remove Singleton

pipermerriam:
## What was wrong?

The `Singleton` class suffered from some shortcomings as well as not quite being general purpose enough.  It also is a common anti-pattern.

## How was it fixed?

Adjusted to an alternate (but similar) approach that isn't constrained to a singleton pattern but rather interning instances based on their class and constructor arguments.  The cache uses a `weakref.WeakValueDictionary` so that we only maintain the cache for objects that are still referenced, automatically cleaning up the cache of any instances that are no longer used.

#### Cute Animal Picture

leopard-predator-animals-eyes-tongue-funny-surprised-face-dw113-living-room-home-wall-art-decor-wood


wasm/_utils/interned.py
```diff
@@ -0,0 +1,28 @@
...
+from weakref import (
+    WeakValueDictionary,
+)
+
+from .caching import (
+    generate_cache_key,
+)
+
+
+class Interned:
```
cburgdorf: I'm still not a fan that each Opcode inherits from `Interned` and therefore bakes in lifecycle management directly. I think I'd prefer something like `registry.get_opcode(OpcodeType, args)` and then the registry would check if such `Opcode` with that args does already exist or not. And the `registry` would just be something that gets instantiated once as part of the bootstrapping. 

But there's probably something that I'm missing (I should re-read what you wrote before but it's almost 1 AM here now) :sleeping:

pipermerriam: I'll wait till Monday before doing anything.

My *justification* for doing it this way is that python internally interns strings and integers.  I don't know the exact mechanism that is used internally to accomplish it so maybe it looks something like this and maybe it doesn't...

Also, this pattern of using `__new__` for stuff isn't without precedent.  I know that `NamedTuple` uses this mechanism for instance creation.

cburgdorf: And I'm not opposed to use that pattern. I personally haven't touched the Singleton pattern (or derivatives) for the last ten years or so and I never *missed* it. But that doesn't mean it doesn't have it's place.

I've been using dependency injection heavily and would probably just require every place to get a `registry` passed in where Opcodes would then be resolved from. However, I do recognize that this may feel clunky to others.

If you feel like reading a bit more about singleton vs DI, this is a good read, too: https://softwareengineering.stackexchange.com/questions/40373/so-singletons-are-bad-then-what

pipermerriam:
Gonna merge and open an issue to re-re-visit this again as the library starts to settle.

PR #40, opened 2019-01-28
Formalize Context, Data Segment, and Element Segment data types

pipermerriam:
Builds on #29 and #41 

## What was wrong?

The context, data segments, element segment objects are all represented using dictionaries.

## How was it fixed?

Implemented formal data types for them.

- https://webassembly.github.io/spec/core/bikeshed/index.html#contexts%E2%91%A0
- https://webassembly.github.io/spec/core/bikeshed/index.html#data-segments%E2%91%A0
- https://webassembly.github.io/spec/core/bikeshed/index.html#element-segments%E2%91%A0

#### Cute Animal Picture

when-you-surprise-an-animal-they-look-like-this-26

### DIFF
--- a/wasm/validation/context.py
+++ b/wasm/validation/context.py
@@ -1,6 +1,6 @@
 from typing import (
-    Optional,
     Tuple,
+    Union,
 )
 
 from wasm.datatypes import (
@@ -25,7 +25,9 @@ class Context:
                  globals: Tuple[GlobalType, ...],
                  locals: Tuple[ValType, ...],
                  labels: Tuple[ValType, ...],
-                 returns: Optional[Tuple[ValType, ...]],
+                 # `returns` may be `None`, but the argument must be
+                 # explicitely supplied.
+                 returns: Union[None, Tuple[ValType, ...]],
                  ) -> None:
         self.types = types
         self.funcs = funcs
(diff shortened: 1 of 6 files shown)


wasm/validation/context.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self,
+                 *,
+                 types: Tuple[FuncType, ...],
+                 funcs: Tuple[FuncType, ...],
+                 tables: Tuple[TableType, ...],
+                 mems: Tuple[MemoryType, ...],
+                 globals: Tuple[GlobalType, ...],
+                 locals: Tuple[ValType, ...],
+                 labels: Tuple[ValType, ...],
+                 returns: Optional[Tuple[ValType, ...]],
```
davesque: Did you forget a default argument value of `None` here?

pipermerriam: No, this was actually intentional.  I want to require the argument, but it can be `None`.  I'll add a comment indicating this.

davesque: Okay, wasn't sure because the `prime` method doesn't require the argument.  But now I'm realizing that the `prime` method doesn't require *any* argument.

pipermerriam: I haven't quite figured out the `Context.prime` API yet.  It's an attempt to remove some/most of the mutation that currently occurs during validation but it's a WIP.

PR #43, opened 2019-01-29
Datatype reorganizatoin

pipermerriam:
Builds on #40 

## What was wrong?

Ended up wanting to re-organize some code.

## How was it fixed?

Reorganized the modules.  This was something that rose out of the WIP in #42 

#### Cute Animal Picture

k04-06-10-31


wasm/tools/fixtures/modules.py
```diff
@@ -49,13 +49,13 @@ def spectest__print(store, arg):
...
         return store, []
 
-    wasm.alloc_func(store, FuncType((ValType.i32,), ()), spectest__print_i32)
-    wasm.alloc_func(store, FuncType((ValType.i64,), ()), spectest__print_i64)
-    wasm.alloc_func(store, FuncType((ValType.f32,), ()), spectest__print_f32)
-    wasm.alloc_func(store, FuncType((ValType.f64,), ()), spectest__print_f64)
-    wasm.alloc_func(store, FuncType((ValType.i32, ValType.f32), ()), spectest__print_i32_f32)
-    wasm.alloc_func(store, FuncType((ValType.f64, ValType.f64), ()), spectest__print_f64_f64)
-    wasm.alloc_func(store, FuncType((), ()), spectest__print)
+    wasm.alloc_func(store, FunctionType((ValType.i32,), ()), spectest__print_i32)
```
davesque: I feel like there was some value in having the naming of things correspond pretty much directly to the naming of entities in the spec doc.  But I don't feel really strongly about it.

pipermerriam: Yeah, I struggled with this a bit too but as things have progressed this felt more natural.  I didn't like `FuncType -> Function -> FunctionInstance` and I didn't like calling `Function -> Func` so I've been defaulting towards the longer form names for things even though it diverges some from the spec.

I suspect that *now*, parity with the spec has been and is valuable, but in a few weeks when this is mostly done, cross referencing the spec will happen less often, and thus will be less important to maintain exact name matching, hence why I'm going to go ahead and :shipit: 

Should be reasonably easy to revert this down the line if I'm wrong or others feel strongly.

PR #44, opened 2019-01-29
Formal data type for module/function/table/memory/global instance types

pipermerriam:
## What was wrong?

The WebAssembly spec defines a formal data type for Module instances:

https://webassembly.github.io/spec/core/bikeshed/index.html#module-instances%E2%91%A0

This type is currently represented using a mutable dictionary.

Within this class exist instance types for memory, table, globals, functions.  Each of these is also represented as a mutable dictionary.

## How was it fixed?

- Implemented the data structures using `NamedTuple` to achieve immutability
- Minor adjustment to the instantiation code to remove circular dependency between function-instance/module-instance/store.
- Implemented formal types for both normal function instances and hostcode function instances.
- implemented formal types for table/memory/global instances.

#### Cute Animal Picture

cone-of-shame-alternatives

### DIFF
--- a/wasm/tools/fixtures/runner.py
+++ b/wasm/tools/fixtures/runner.py
@@ -16,6 +16,9 @@ from _pytest.outcomes import (
     Failed,
 )
 import wasm
+from wasm.datatypes import (
+    ModuleInstance,
+)
 from wasm.exceptions import (
     Exhaustion,
     InvalidModule,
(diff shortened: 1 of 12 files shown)


wasm/tools/fixtures/runner.py
```diff
@@ -79,9 +82,9 @@ def instantiate_module_from_wasm_file(
 
             sub_module = registered_modules[import_.module]
 
-            for export in sub_module["exports"]:
+            for export in sub_module.exports:
                 if export.name == import_.name:
-                    externval = export.desc
+                    externval = export.value
```
pipermerriam: This was me incorrectly using the `Export` type instead of the `ExportInstance` type.

PR #45, opened 2019-01-29
Formalize Configuration object and execution data structures

pipermerriam:
## What was wrong?

The `config` object used for all execution by the wasm interpreter is just a mutable dictionary with multiple mutable lists that are manipulated.

## How was it fixed?

- Implemented a `Configuration` class which houses the three stacks (values, labels, frames) and the store.
- Instruction sequences are now housed in their respective frames and labels.
- A new `InstructionSequence` class for managing the state of the stream of instructions.  This has removed the need for the `"continuation"` fields in label and frame because the parent instruction sequences automatically keep track of their position.
- New `Stack` data structures with `push()` and `pop()` APIs.

> Note:  This PR noticeably decreased performance (tests take longer to execute).  I'm planning to address this later.

#### Cute Animal Picture

e-collar-blog

### DIFF
--- a/wasm/datatypes/configuration.py
+++ b/wasm/datatypes/configuration.py
@@ -1,0 +1,62 @@
+from typing import (
+    NamedTuple,
+)
+
+from wasm.typing import (
+    Store,
+)
+
+from .instructions import (
+    InstructionSequence,
+)
+from .stack import (
+    Frame,
+    FrameStack,
+    Label,
+    LabelStack,
+    ValueStack,
+)
+
+
+class Configuration(NamedTuple):
+    store: Store
+    value_stack: ValueStack
+    label_stack: LabelStack
+    frame_stack: FrameStack
+
+    @property
+    def frame(self) -> Frame:
+        return self.frame_stack.peek()
+
+    @property
+    def has_active_frame(self):
+        return bool(self.frame_stack)
+
+    @property
+    def label(self) -> Label:
+        if self.has_active_frame and self.has_active_label:
+            return self.label_stack.peek()
+        else:
+            raise IndexError("No labels active for current frame")
+
+    @property
+    def has_active_label(self):
+        if not self.has_active_frame:
+            return False
+        elif not self.label_stack:
+            return False
+        else:
+            return self.label_stack.peek().frame_id == self.frame.id
+
+    @property
+    def instructions(self) -> InstructionSequence:
+        if self.has_active_label:
+            return self.label.instructions
+        elif len(self.frame_stack):
+            return self.frame_stack.peek().instructions
+        else:
+            raise ValueError("No active instructions")
+
+    @property
+    def is_executable(self) -> bool:
+        return self.has_active_frame
--- a/wasm/datatypes/instructions.py
+++ b/wasm/datatypes/instructions.py
@@ -1,0 +1,69 @@
+from collections.abc import (
+    Sequence,
+)
+from typing import (
+    TYPE_CHECKING,
+    Iterator,
+    Tuple,
+    overload,
+)
+
+if TYPE_CHECKING:
+    from wasm.instructions import (  # noqa: F401
+        BaseInstruction,
+    )
+
+
+class InstructionSequence(Sequence):
+    _instructions: Tuple['BaseInstruction', ...]
+
+    def __init__(self, instructions: Tuple['BaseInstruction', ...]) -> None:
+        self._instructions = instructions
+        self._idx = -1
+
+    def __str__(self) -> str:
+        return f"[{' > '.join((str(instr) for instr in self._instructions))}]"
+
+    def __repr__(self) -> str:
+        return f"InstructionSequence({str(self)})"
+
+    def __len__(self) -> int:
+        return len(self._instructions)
+
+    def __next__(self) -> 'BaseInstruction':
+        self._idx += 1
+        try:
+            return self._instructions[self._idx]
+        except IndexError:
+            raise StopIteration
+
+    def __iter__(self) -> Iterator['BaseInstruction']:
+        while self._idx < len(self._instructions) - 1:
+            yield next(self)
+
+    @overload
+    def __getitem__(self, idx: int) -> 'BaseInstruction':
+        pass
+
+    @overload  # noqa: 811
+    def __getitem__(self, s: slice) -> 'InstructionSequence':
+        pass
+
+    def __getitem__(self, item):  # noqa: F811
+        if isinstance(item, int):
+            return self._instructions[item]
+        elif isinstance(item, slice):
+            return type(self)(self._instructions[item])
+        else:
+            raise TypeError(f"Unsupported key type: {type(item)}")
+
+    @property
+    def pc(self) -> int:
+        return max(0, self._idx)
+
+    @property
+    def current(self) -> 'BaseInstruction':
+        return self[self.pc]
+
+    def seek(self, idx: int) -> None:
+        self._idx = idx - 1
--- a/wasm/datatypes/stack.py
+++ b/wasm/datatypes/stack.py
@@ -1,0 +1,86 @@
+from typing import (
+    Generic,
+    Iterable,
+    List,
+    NamedTuple,
+    Tuple,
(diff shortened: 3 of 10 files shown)


wasm/datatypes/configuration.py
```diff
@@ -0,0 +1,62 @@
...
+    def has_active_label(self):
+        if not self.has_active_frame:
+            return False
+        elif not self.label_stack:
+            return False
+        else:
+            return self.label_stack.peek().frame_id == self.frame.id
+
+    @property
+    def instructions(self) -> InstructionSequence:
```
pipermerriam: I'm not confident in this API but it's working right now.  My spider sense says:

- These `property` methods are expensive
- The frame/label juggling may be error prone or it may only work incidentally but not actually be truly spec compliant.

wasm/datatypes/stack.py
```diff
@@ -0,0 +1,86 @@
...
+class FrameStack(BaseStack[Frame]):
+    pass
+
+
+class Label(NamedTuple):
+    arity: int
+    height: int
+    instructions: InstructionSequence
+    is_loop: bool
+    frame_id: uuid.UUID
```
pipermerriam: This is one of the things that isn't in the spec but that I needed to implement for the frame/stack juggling to work correctly.  Basically, when a new `Frame` lands on the frame stack, the label stack should be cleared.  However, once the new frame exits and the old frame resumes, the old label stack needs to be restored.

I accomplished this by tieing the labels to the frame their were created in allowing us to treat the label stack as empty if the top label's frame isn't the top frame.

wasm/main.py
```diff
@@ -2286,94 +2263,90 @@ def spec_invoke_function_address(config, a=None):
...
-                "arity": len(t2m),
-                "height": len(operand_stack),
-                "continuation": [instrstar, idx + 1, control_stack],
-            }
-        ]
-        # 12
-        blockinstrstarendend = [Block(t2m, tuple(instrstarend)), End()]
-        config["instrstar"] = blockinstrstarendend
-        config["idx"] = 0
-        config["control_stack"] = []
```
pipermerriam: Here's the part I was referring to about clearing the label stack.  See above how the `"continuation"` contains the old label stack (previously called control_stack).  This no longer needs to be explicitly tracked like this, however, the code doesn't currently provide strong guarantees that we can't get into a situation where there is a label on the label stack which refers to a frame that is no longer part of the frame stack.

wasm/main.py
```diff
@@ -2569,21 +2542,26 @@ def instrstarend_loop(config):
...
+        #logger.debug('INSTRUCTION: %s', instruction.opcode.text)
+        #logger.debug('LOCALS: %s', config.frame.locals)
+        #logger.debug('NUM-VALUES: %d', len(config.value_stack))
+        #logger.debug('NUM-LABELS: %d', len(config.label_stack))
+        #logger.debug('NUM-FRAMES: %d', len(config.frame_stack))
+        #logger.debug('VALUES: %s', tuple(config.value_stack))
+
         logic_fn = opcode2exec[instruction.opcode][0]
-        ret = logic_fn(config)
+        logic_fn(config)
```
pipermerriam: This section is probably one of the larger functionality changes.  (Note that this needs to be seen in the context of the changes to `spec_end`).

Previously, execution exited once a logic function returned a truthy value.  This was a very unreliable mechanism because **any** logic function that chose to return a truthy value would have resulted in the interpreter thinking that it was done.

The new mechanism *should* correctly continue execution until it exhausts the sequence of instructions, returning whatever value is hanging out on the stack when that is done.

wasm/datatypes/stack.py
```diff
@@ -0,0 +1,86 @@
...
+class Frame(NamedTuple):
+    id: uuid.UUID
+    module: ModuleInstance
+    locals: List[TValue]
+    instructions: InstructionSequence
+    arity: int
+    height: int
+
+
+class FrameStack(BaseStack[Frame]):
```
cburgdorf: Is there a reason why this needs to be a class? A simple alias `FrameStack = BaseStack[Frame]` should do it unless there's something I'm not aware of.

cburgdorf: Ah, I see the `LabelStack` further down adds a method, so maybe the class based syntax is just for consistency then. Feel free to ignore.

wasm/datatypes/configuration.py
```diff
@@ -0,0 +1,62 @@
...
+    LabelStack,
+    ValueStack,
+)
+
+
+class Configuration(NamedTuple):
+    store: Store
+    value_stack: ValueStack
+    label_stack: LabelStack
+    frame_stack: FrameStack
```
cburgdorf: Should all of these be named with underscore prefix to maintain pseudo privacy?

wasm/datatypes/configuration.py
cburgdorf: >These `property` methods are expensive

You mean compared to simple methods? According to my `timeit` a measurable performance different comes only into play for the *instance creation*  of `Configuration` because that is where the `@property` decorator syntax is evaluated. Once, the instance is created, there is no measurable difference between calling a *property*  or a *method*.

In other words, considering the performance implication of the `@property` syntax should only matter if we happen to create lots of `Configuration` instances repeatedly (which I doubt is happening, is it?)

wasm/datatypes/instructions.py
```diff
@@ -0,0 +1,69 @@
...
+    def __getitem__(self, item):  # noqa: F811
+        if isinstance(item, int):
+            return self._instructions[item]
+        elif isinstance(item, slice):
+            return type(self)(self._instructions[item])
+        else:
+            raise TypeError(f"Unsupported key type: {type(item)}")
+
+    @property
+    def pc(self) -> int:
```
cburgdorf: What does `pc` stand for?

ralexstokes: 'program counter', like 'instruction counter'; an index into the `InstructionSequence`. this is how the execution engine knows "what is the next instruction to execute"

cburgdorf: I feel like that would make a better name for it ;)

pipermerriam: https://github.com/ethereum/py-evm/blob/master/eth/vm/code_stream.py#L60-L62

I just defaulted to the same name used in the EVM.  I'll add a docstring to help clarify what this value represents.

wasm/datatypes/stack.py
```diff
@@ -0,0 +1,86 @@
...
+class Label(NamedTuple):
+    arity: int
+    height: int
+    instructions: InstructionSequence
+    is_loop: bool
+    frame_id: uuid.UUID
+
+
+class LabelStack(BaseStack[Label]):
+    def get_by_label_idx(self, key: LabelIdx) -> Label:
```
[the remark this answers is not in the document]
pipermerriam: Used here: https://github.com/ethereum/py-wasm/pull/45/files#diff-8e77ec4ec9a48540b36b24e9a6786161R2097

Probably missed because that diff is collapsed by default.

wasm/datatypes/configuration.py
pipermerriam: I think you missed the `main.py` module diff (which is collapsed by default)  These are used heavily as a public API on the `config` object.

wasm/datatypes/configuration.py
pipermerriam: I'll profile this but the performance hit comes from the inner logic of checking which frame or label it should get the instruction sequence from.   In the previous implementation this was just a mutable property that was continually overwritten with whatever the active instruction sequence was.

wasm/datatypes/stack.py
pipermerriam: I had similar thoughts.

The `peek()`, `pop2()` and `pop3()` APIs are all used and losing them isn't a big deal.  I'll ponder a bit but I'm leaning towards keeping them as formal data structures.

wasm/datatypes/configuration.py
cburgdorf: Ah! That's why!

PR #46, opened 2019-01-31
Implement formal datatype for store

pipermerriam:
Builds on #45

## What was wrong?

The Web Assembly spec defines a "Store" which is a sort of global environment for wasm execution.

## How was it fixed?

Implemented the data structure using a `NamedTuple`.

This PR also contains the following cleanups:

1. Found and cleaned up one location where strings were still being used to represent types (in the type validation of arguments being passed into the wasm interpreter)
2. Removed multiple un-used and broken `spec_<thing>_inv` functions that had no test coverage and were not used (and won't be used because I'll be writing all of these new as stream based [serializers/composers/unparsers](https://stackoverflow.com/questions/148857/what-is-the-opposite-of-parse)
3. Enable full `flake8` linting on the `wasm.main` module.
4. Lots of cleanups for exception messages and broken f-strings (detected by linting)
5. Further filling in of missing type hints.

#### Cute Animal Picture

cone_of_shame

### DIFF
--- a/wasm/datatypes/configuration.py
+++ b/wasm/datatypes/configuration.py
@@ -1,0 +1,62 @@
+from typing import (
+    NamedTuple,
+)
+
+from wasm.typing import (
+    Store,
+)
+
+from .instructions import (
+    InstructionSequence,
+)
+from .stack import (
+    Frame,
+    FrameStack,
+    Label,
+    LabelStack,
+    ValueStack,
+)
+
+
+class Configuration(NamedTuple):
+    store: Store
+    value_stack: ValueStack
+    label_stack: LabelStack
+    frame_stack: FrameStack
+
+    @property
+    def frame(self) -> Frame:
+        return self.frame_stack.peek()
+
+    @property
+    def has_active_frame(self):
+        return bool(self.frame_stack)
+
+    @property
+    def label(self) -> Label:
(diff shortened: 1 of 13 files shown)


wasm/datatypes/configuration.py
```diff
@@ -0,0 +1,61 @@
...
+    Label,
+    LabelStack,
+    ValueStack,
+)
+from .store import (
+    Store,
+)
+
+
+class Configuration(NamedTuple):
```
davesque: Some concerns about this class:

* Its definition and function seem to diverge a bit from the spec.  Namely, configurations in the spec are just a tuple of references to some current *store*, *frame*, and *instruction sequence*.  Then all the instructions essentially act on a mutable configuration.
* I think the separation of the execution stack into three homogeneously typed stacks might be problematic.  This might account for some of the difficulty in tracking which labels belong to which frame.  In terms of the semantics defined by the spec, I believe it's correct to say that all labels that appear immediately above a frame in the execution stack belong to that frame.  It's then easy to pop off a frame's labels; you just pop until you hit a frame.  Also, the semantics of the reduction rules are, I think, fairly well-coupled with the notion of a single, heterogeneous stack.

The above concerns are tied in with my intuition that it will be very useful if the implementation can correspond as much as possible with the spec.  That should help new developers a lot.

PR #52, opened 2019-02-01
Convert remaining binary parser to be stream based

pipermerriam:
## What was wrong?

There was still a large swath of legacy code for parsing binary modules that did index tracking into the bytestream.

## How was it fixed?

Finish replacing all parsers with stream based parsers.  This also fixes two spec failures that were due to incorrect parsing in the legacy code.

#### Cute Animal Picture

funny-pet-cones-creative-elizabethan-collars-1


wasm/parsers/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    parse_module,
+)
+from .mutability import (  # noqa: F401
+    parse_mut,
+)
+from .sections import (  # noqa: F401
+    parse_sections,
+    parse_custom_section,
+    parse_function_section,
+    parse_import_section,
```
pipermerriam: Most of these imports can be removed as they don't need to be public APIs.

pipermerriam:
This one is quite large, however it's all quite mundane replacing the previous binary parsers with stream based parsers, replicating the logic 1:1 with the exception of module parsing which now *correctly* implements the parsing algorithm (which you can see in the two expected failures that have been removed from `runner.py`)

I can pull this apart if anyone feels strongly but I think it will be adequate to get two people's :eyes: to sanity check this.  The spec tests give me a high confidence in the correctness of this PR.

wasm/tools/fixtures/runner.py
```diff
@@ -126,13 +123,8 @@ def run_opcode_action(command, store, module, all_modules, registered_modules):
...
     # must be converted to unicode string
-    funcname = action.field
-    idx = 0
-    utf8_bytes = bytearray()
-    for c in funcname:
-        utf8_bytes += bytearray([ord(c)])
-    utf8_bytes = wasm.spec_binary_uN_inv(len(funcname), 32) + utf8_bytes
-    _, funcname = wasm.spec_binary_name(utf8_bytes, 0)
+    raw_funcname = action.field
+    funcname = bytes(bytearray([ord(c) for c in raw_funcname])).decode('utf8')
```
carver: I think this is the same thing as:

```py
funcname = raw_funcname.encode('ascii').decode('utf8')
```

wasm/tools/fixtures/runner.py
```diff
@@ -340,18 +332,6 @@ def get_command_fn(command):
...
-        # allowing the parser to exit before fully parsing the module.  For
-        # this test it ends up exiting before encountering the invalid section
-        # which **should** error out due to an invalid section ID.
-        93: Failed,
-        # This one appears to be lack of validation of the lengths of code and
-        # function section types.
-        # See the note here about malformed modules:
-        # - https://webassembly.github.io/spec/core/bikeshed/index.html#code-section%E2%91%A0
-        102: Failed,
-    },
```
carver: :tada: :tada:

wasm/parsers/sections.py
```diff
@@ -0,0 +1,486 @@
...
+            if section_id == parser_section_id:
+                yield section_parser(stream)
+                # Track which section ids we've encountered.
+                seen_section_ids.add(section_id)
+                break
+            else:
+                yield EMPTY_SECTIONS_BY_ID[parser_section_id]
+        else:
+            # If the loop above exits naturally, it means that we've
+            # encountered a section out of order or multiple times.
```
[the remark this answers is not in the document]
pipermerriam: This looks more intuitively grokable, will update to something loosely shaped like this.

wasm/parsers/version.py
```diff
@@ -0,0 +1,34 @@
...
+from wasm.typing import (
+    UInt8,
+)
+
+from .byte import (
+    parse_single_byte,
+)
+
+KNOWN_VERSIONS = {
+    (0x01, 0x00, 0x00, 0x00),
```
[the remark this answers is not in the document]
pipermerriam: I don't think that's correct.  Have a look at this.

https://webassembly.github.io/spec/core/bikeshed/index.html#modules%E2%91%A0%E2%93%AA

That seems pretty clear that the 3 empty bytes are part of the version and not a custom section.

wasm/tools/fixtures/runner.py
pipermerriam: So much better.  :+1:

wasm/tools/fixtures/runner.py
pipermerriam: Yeah, this felt like really nice validation that the new structure is *correct*.  When I finished implementing the stream parser and unwrapped a few subtle transcription errors, these just started failing (i.e. passing when they were expected failures).

wasm/parsers/module.py
```diff
@@ -0,0 +1,66 @@
...
+    parse_sections,
+)
+from .version import (
+    parse_version,
+)
+
+
+def parse_module(stream: io.BytesIO) -> Module:
+    parse_magic(stream)
+    parse_version(stream)
```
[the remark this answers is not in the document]
pipermerriam: :+1: I think I'll probably go with docstring/comments to explain why `magic` is discarded and will add an extra field to the module to store the version so that we can apply a `validate_version` to it during validation.

wasm/parsers/version.py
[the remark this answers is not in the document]
carver: Why not use the same encoding for integer that they use everywhere else? Such a strange choice by the wasm spec, but yeah, the trailing bytes are clearly indicated.

PR #55, opened 2019-02-03
Isolate mutation during expression validation

### LINKED ISSUE #54: Remove vague mutability from validation API
### Probem?

Right now the `wasm.validation.context.Context` object houses a wide array of validation code. The *static* validation methods like `validate_label_idx` are acceptable as they don't mutate.

However, expression validation is inherently mutation based.  As you step through the opcodes you maintain a control stack and operand-type stack, validating for each function that there will be the properly typed values present on the stack.

The way this currently works is the `Context` object exposes methods for interacting with the validation stacks and these methods do internal mutation.  Were a context object to be re-used across validating multiple expressions, it would be easy for the results to be wrong.

### Fix?

Split this into two APIs.

- `Context`: static data container
- `ExpressionContext`: wraps a `Context`, has a defined lifecycle, contains mutation

```python
context = Context(...)

# consider maintaining the simple validation APIs
context.validate_label_idx(...)

# for situations where we need to extend/modify the context
with context.copy(<overrides>) as ctx:
    ...

# for expression validation
with context.validate_expression(<overrides>) as something:
    ...
```

In the above pseudo code, the `something` variable would have the various `push_operand()` APIs necessary to do expression validation.  In addition, it will have a well defined lifecycle.  Once an expression has *returned* the object becomes decommissioned, disallowing any of the mutation based APIs as well as exposing the return types to the expression which would not be available until the object has been decommissioned.

pipermerriam:
builds on #52 

fixes #54 

## What was wrong?

As described in #54 expression validation involved mutating the `Context` object.  This has some unfortunate ways this can fail.

## How was it fixed?

Adjust API for much stronger encapsulation of the mutable components.

#### Cute Animal Picture

creative-pet-cones-elizabethan-collars-3__605

### DIFF
--- a/wasm/instructions/control.py
+++ b/wasm/instructions/control.py
@@ -48,6 +48,18 @@ class Block(NamedTuple):
         rt = f"({','.join((v.value for v in self.result_type))})"
         return f"{self.opcode.text}[rt={rt},expr={stringify_instructions(self.instructions)}]"
 
+    @classmethod
+    def wrap(self,
+             result_type: Tuple[ValType, ...],
+             expression: Tuple[BaseInstruction, ...]
+             ) -> Tuple[BaseInstruction, ...]:
+        return cast(Tuple[BaseInstruction, ...], (
+            Block(
+                result_type,
+                expression,
+            ),
+        ))
+
 
 @register
 class Br(Interned):
--- a/wasm/instructions/control.py
+++ b/wasm/instructions/control.py
@@ -49,12 +49,12 @@ class Block(NamedTuple):
         return f"{self.opcode.text}[rt={rt},expr={stringify_instructions(self.instructions)}]"
 
     @classmethod
-    def wrap(self,
+    def wrap(cls,
              result_type: Tuple[ValType, ...],
              expression: Tuple[BaseInstruction, ...]
              ) -> Tuple[BaseInstruction, ...]:
         return cast(Tuple[BaseInstruction, ...], (
-            Block(
+            cls(
                 result_type,
                 expression,
             ),
--- a/wasm/validation/context.py
+++ b/wasm/validation/context.py
@@ -1,4 +1,5 @@
 from typing import (
+    Optional,
     Tuple,
     Union,
 )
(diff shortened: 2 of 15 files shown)


pipermerriam:
only 6819015 is relevent for this PR

wasm/instructions/control.py
```diff
@@ -48,6 +48,18 @@ def __str__(self) -> str:
         rt = f"({','.join((v.value for v in self.result_type))})"
         return f"{self.opcode.text}[rt={rt},expr={stringify_instructions(self.instructions)}]"
 
+    @classmethod
+    def wrap(self,
```
carver: *nit* `wrap(cls,`

wasm/instructions/control.py
```diff
@@ -48,6 +48,18 @@ def __str__(self) -> str:
...
+             result_type: Tuple[ValType, ...],
+             expression: Tuple[BaseInstruction, ...]
+             ) -> Tuple[BaseInstruction, ...]:
+        return cast(Tuple[BaseInstruction, ...], (
+            Block(
```
carver: Maybe `cls(` here

wasm/validation/context.py
```diff
@@ -35,7 +36,110 @@
...
+            )
+
+    def has_local(self, idx: LocalIdx) -> bool:
+        return idx < len(self.locals)
+
+    def get_local(self, idx: LocalIdx) -> ValType:
+        return self.locals[idx]
+
+
+class Context(BaseContext):
```
carver: Alternative name brainstorm:
- `Inventory`
- ~`Register`~
- ~`Record`~
- `Lookup`
- ~`Index`~
- ~`Tabulation`~
- `Catalog`

I'm focusing on the idea that it's sort of a complete register of all items, and a mechanism to look them up. A few are struck out for being too heavily used in other places, or confused for verbs.

wasm/validation/expressions.py
```diff
@@ -49,27 +49,22 @@ def validate_expression(expression: Tuple[BaseInstruction, ...],
 
         logger.debug('Validating instruction: %s', instruction)
 
-        # SIDE-EFFECT: Instruction validation is inherently
-        # stateful/side-effect causing.  Both `Context.operand_stack` and
```
carver: Do I have this right?

This method is still causing side-effects, but we don't care because we made a defensive copy to the `ExpressionContext` first.

wasm/validation/function.py
```diff
@@ -1,13 +1,60 @@
...
+    )
+    with ctx:
+        # validate body using algorithm in appendix
+        instrstar = cast(Tuple[BaseInstruction, ...], (
+            Block(
+                function_type.results,
+                function.body,
+            ),
+        ))
+        validate_expression(instrstar, ctx)
```
carver: What does `instrstar` mean? Does `expression` work just as well?

carver:
> Things I'm not happy about:
> 
> * I don't love how `Context` and `ContextExpression` are structured
> * The name `ExpressionContext`
> * The name of the `Context.expression_context()` API
> 
> They all feel clunky.

Hm, no major suggestions. Just realized you were asking about `ExpressionContext` instead of `Context`. Is it reasonable to call it an execution context? At this point it's all used for validation. Maybe `ValidationContext`?

wasm/validation/function.py
pipermerriam: Yeah, leftover variable names.  Updated to `expression`.

PR #56, opened 2019-02-03
Normalize object/function/variable names for *function* things.

pipermerriam:
Builds on #55

## What was wrong?

The WebAssembly spec uses the name `func` for most references to function *things*.  This is not 100% consistent as things like [start function](https://webassembly.github.io/spec/core/bikeshed/index.html#start-function%E2%91%A0) use the full form.

As the codebase has evolved:

- The `func` abbreviation felt like it was negatively effecting readability.(`func_idx`, `func_addr`, `func_type`, `func`, `hostfunc`
- Cases like the start function use the full form `function`.

While spec conformance by using the name `func` is likely to have some benefits, i.e. easier mapping between the spec and the codebase, I'm not convinced this outweighs the readability improvements that come with using the full form name.  Also, by using the full form name everywhere it became easier to intuitively know what name something would use, where-as before, there was some inconsistency between APIs, such as `Export.is_function` which one might think would be `Export.is_func` if we were using a mix of short-form and long-form.

## How was it fixed?

Converted *almost* all of these to use the long form.  The exceptions which I still plan to cleanup are some of the container types like `Store` and `Module`, both of which still use the short form name.

#### Cute Animal Picture

cone-of-shame-18-flower

### DIFF
--- a/wasm/main.py
+++ b/wasm/main.py
@@ -25,9 +25,9 @@ from wasm._utils.validation import (
 from wasm.datatypes import (
     Export,
     ExportInstance,
-    FuncIdx,
     Function,
     FunctionAddress,
+    FunctionIdx,
     FunctionInstance,
     FunctionType,
     GlobalAddress,
@@ -45,7 +45,6 @@ from wasm.datatypes import (
     Module,
     ModuleInstance,
     Mutability,
-    StartFunction,
     Store,
     TableAddress,
     TableIdx,
@@ -116,6 +115,7 @@ from wasm.validation import (
     validate_global,
     validate_memory,
     validate_memory_type,
+    validate_start_function,
     validate_table,
     validate_table_type,
 )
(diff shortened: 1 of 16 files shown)


pipermerriam:
only 7f15993 is relevant for this PR.

wasm/main.py
```diff
@@ -2376,8 +2365,8 @@ def spec_allocfunc(S: Store,
     logger.debug('spec_allocfunc()')
 
     funcaddr = FunctionAddress(len(S.funcs))
-    func_type = module.types[func.type_idx]
-    funcinst = FunctionInstance(func_type, module, func)
+    function_type = module.types[func.type_idx]
+    funcinst = FunctionInstance(function_type, module, func)
```
carver: Looks like all the other `func` variables inside this method could be updated as well.

pipermerriam: To avoid the merge conflict I'm going to leave as is since this gets cleaned up in https://github.com/ethereum/py-wasm/pull/59/files#diff-26f7706e7df969dc490d122a6f78ae05R55

PR #62, opened 2019-02-04
[WIP] Piper/formalize execution environment

pipermerriam:
## What was wrong?

Issue #

## How was it fixed?

Summary of approach.

#### Cute Animal Picture


wasm/datatypes/store.py
```diff
@@ -26,19 +32,49 @@
...
 from .table import (
     TableInstance,
     TableType,
 )
 
+TAddress = Union[FunctionAddress, GlobalAddress, MemoryAddress, TableAddress]
+TExtern = Union[FunctionType, TableType, MemoryType, GlobalType]
+
+
+def _collate_addresses(exports: Tuple[Export, ...],
```
pipermerriam: rename to `_collate_exports`

wasm/datatypes/store.py
```diff
@@ -52,6 +88,46 @@ def __init__(self) -> None:
         self.tables = []
         self.globals = []
 
+    def get_type(self, address: TAddress) -> TExtern:
```
pipermerriam: rename to `get_type_for_address`

wasm/datatypes/store.py
```diff
@@ -85,8 +179,94 @@ def allocate_memory(self, memory_type: MemoryType) -> MemoryAddress:
...
+        module_instance = ModuleInstance(
+            types=module.types,
+            func_addrs=function_addresses,
+            table_addrs=table_addresses,
+            memory_addrs=memory_addresses,
+            global_addrs=global_addresses,
+            exports=exports,
+        )
+
+        # We have to perform the actualf function allocations *after* module
```
pipermerriam: s/actualf/actual

wasm/execution/runtime.py
```diff
@@ -0,0 +1,256 @@
...
+            raise Unlinkable(
+                f"No export found with name '{import_.name}'"
+            )
+
+
+def _initialize_globals(store: Store,
+                        module: Module,
+                        globals_addresses: Tuple[GlobalAddress, ...],
+                        ) -> Iterator[UInt32]:
+        # 5
```
pipermerriam: remove

wasm/execution/runtime.py
```diff
@@ -0,0 +1,256 @@
...
+        ))
+        data_segment_offsets = tuple(_compute_data_offsets(
+            self.store,
+            module.data,
+            module_instance,
+        ))
+
+        # 11
+        # 12
+        # 13
```
pipermerriam: remove

wasm/execution/runtime.py
```diff
@@ -0,0 +1,256 @@
...
+                table_instance = self.store.tables[table_address]
+                table_instance.elem[offset + idx] = function_address
+
+        for offset, data_segment in zip(data_segment_offsets, module.data):
+            memory_address = module_instance.memory_addrs[data_segment.memory_idx]
+            memory_instance = self.store.mems[memory_address]
+            data_length = len(data_segment.init)
+            memory_instance.data[offset:offset + data_length] = data_segment.init
+
+        # 15
```
pipermerriam: remove

pipermerriam:
Only 5f8cede is relevant to this PR.

carver:
> Only [5f8cede](https://github.com/ethereum/py-wasm/commit/5f8cede5d29aa44c134e4342a3d7d00727b7efc4) is relevant to this PR.

Presumably dd0bc61cf44 instead? These stacked PRs are getting pretty bulky. I would hate to miss reviewing a commit because I was looking at the wrong one. Maybe I'll wait for the others to merge and this to rebase.

wasm/execution/configuration.py
```diff
@@ -125,6 +129,23 @@ def __init__(self, store: Store) -> None:
...
         self.frame_stack = FrameStack()
 
+    def execute(self) -> Tuple[TValue, ...]:
+        # TODO: unwrap import once logic functions move out of wasm.main
+        from wasm.main import opcode2exec
+
+        while self.has_active_frame:
+            instruction = next(self.instructions)
+
+            logic_fn = opcode2exec[instruction.opcode][0]
```
carver: I can't remember why there's a `[0]` at the end here, or what is in the later indices. Sorry, no suggestions, just identifying a part that wasn't immediately apparent.

wasm/execution/configuration.py
```diff
@@ -125,6 +129,23 @@ def __init__(self, store: Store) -> None:
...
+            logic_fn(self)
+
+        if len(self.result_stack) > 1:
+            raise Exception("Invariant?")
```
carver: Heh, maybe a bit more info? Could be length of the stack, items on it, etc.

wasm/execution/runtime.py
```diff
@@ -0,0 +1,251 @@
...
+            frame = Frame(
+                module=module_instances,
+                locals=[],
+                instructions=InstructionSequence(global_.init),
+                arity=1,
+            )
+            config.push_frame(frame)
+            result = config.execute()
+            if len(result) != 1:
+                raise Exception("Invariant: globals initialization returned empty result")
```
carver: Didn't you add a `get_single_or_crash()` kind of thing somewhere?

wasm/tools/fixtures/runner.py
```diff
@@ -65,34 +78,19 @@ def instantiate_module_from_wasm_file(
...
-        # validate
-        wasm.validate_module(module)
+    wasm.validate_module(module)
 
-        # imports preparation
-        externvalstar: List[Any] = []
-        for import_ in module.imports:
-            if import_.module not in registered_modules:
-                raise Unlinkable(f"Unlinkable module: {import_.module}")
+    module_instance, _ = runtime.instantiate_module(module)
```
carver: Sweet. I guess none of the tests are validating the return value form the loading the module? If they're all returning `None`, maybe we can just assert that.

wasm/tools/fixtures/runner.py
```diff
@@ -276,52 +331,69 @@ def do_assert_unlinkable(command, store, module, all_modules, registered_modules
...
+        runtime.register_module(command.as_, all_modules[command.name])
 
 
-def do_action_command(command, store, module, all_modules, registered_modules):
-    run_opcode_action(command, store, module, all_modules, registered_modules)
+def do_action_command(command: TActionCommands,
+                      module: CurrentModule,
+                      all_modules: AllModules,
+                      runtime: Runtime) -> None:
+    run_opcode_action(command, module, all_modules, runtime)
```
carver: It confused me that this method exists. I guess just for the `do_*` consistency? It made me wonder if I was missing something else it did. Maybe it can be replaced with an alias, like:
```py
do_action_command = run_opcode_action
```

wasm/execution/configuration.py
pipermerriam: This is the old/legacy implementation still bleading through.  It's setup as `Dict[BinaryOpcode, Tuple[Callable, ...]]`.  Most of the opcodes are a single function (thus the `[0]` indexing).  The numeric operations are done as two callables, one that handles popping values off the stack and then passing the values into the second function.

This pattern is going away with each opcode just mapping to one logic function.

wasm/execution/configuration.py
pipermerriam: Updated to make it clear that the WASM spec only allows at most a single return value.  This is done as an `Invariant` using raw `Exception` because validation **should** always catch this before we could even reach this, so hitting this would indicate that there is something wrong with our implementation.

wasm/execution/runtime.py
pipermerriam: Yes, but it's a bit more awkward to implement in a generic way since the error messages need to be different in each case.  Similar to the other one, this is a runtime check which we should never be able to hit since this would be caught during validation.

wasm/tools/fixtures/runner.py
pipermerriam: I don't think they all return `None`, it's just that we discard this value because the spec tests test this using a level of indirection.  The spec tests load a module with a start function which sets one of the globals to a value and the tests assert that the value is set to the right value.  

I think it's still appropriate for the instantiation API to return the result value from the start function, but in this case, during spec test runs I don't think there's much we can do with the value.

PR #68, opened 2019-02-07
caching for opcode.text property

pipermerriam:
## What was wrong?

The `opcode.text` property involved a bit of lookup overhead.

## How was it fixed?

Cache the value the first time it is accessed.

#### Cute Animal Picture

382231e5aae0ea9dbf8a2b50a5640961


pipermerriam:
It was a false start at profiling.  I realized  I need to finish wiping out the `wasm.main` module before I can really touch profiling and performance.

PR #69, opened 2019-02-07
signed and unsigned conversion utils and tests

pipermerriam:
## What was wrong?

We only had `unsigned > signed` conversions.

## How was it fixed?

Fills out the `signed > signed` and adds round trip hypothesis tests to increase confidence of correctness.

#### Cute Animal Picture

cone_of_shame_640x4201

### DIFF
--- a/tests/core/numeric-utils/test_unsigned_to_signed.py
+++ b/tests/core/numeric-utils/test_unsigned_to_signed.py
@@ -1,0 +1,57 @@
+from hypothesis import (
+    given,
+    strategies as st,
+)
+import pytest
+
+from wasm import (
+    constants,
+)
+from wasm._utils.numeric import (
+    u32_to_s32,
+    u64_to_s64,
+)
+
+
+@pytest.mark.parametrize(
+    'convert_fn,value,expected',
+    (
+        (u32_to_s32, 0, 0),
+        (u64_to_s64, 0, 0),
+        (u32_to_s32, 1, 1),
+        (u64_to_s64, 1, 1),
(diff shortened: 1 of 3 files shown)


tests/core/numeric-utils/test_unsigned_to_signed.py
```diff
@@ -0,0 +1,57 @@
...
+
+@pytest.mark.parametrize(
+    'convert_fn,value,expected',
+    (
+        (u32_to_s32, 0, 0),
+        (u64_to_s64, 0, 0),
+        (u32_to_s32, 1, 1),
+        (u64_to_s64, 1, 1),
+        (u32_to_s32, constants.SINT32_CEIL, constants.SINT32_MIN),
+        (u64_to_s64, constants.SINT64_CEIL, constants.SINT64_MIN),
```
carver: I guess the value just below the ceiling is worth checking, too. Like: `(u32_to_s32, constants.SINT32_CEIL - 1, constants.SINT32_CEIL - 1)`

pipermerriam: These functions don't validate inputs so their behavior isn't defined if the input value is out of range.  Putting in the runtime checks adds overhead that I don't think we can allow.

The things that I think make this ok are:

- Any values that come from parsing a binary module will have bounds checked.
- Any user supplied input values will have bounds checking done on the way in.

This still allows for a logic function to behave incorrectly and to not properly wrap overflowing values or something similar but the spec tests cover at least some of that surface area.

PR #70, opened 2019-02-07
ValTyp argument and zero value APIs

pipermerriam:
## What was wrong?

Two APIs that are needed for incoming functionality.

- Ability to validate that a value is appropriate for a given value type.
- Ability to generate the ZERO value for a given value type.

## How was it fixed?

Two new APIs on the `ValType`

- `ValType.zero_value`
- `ValType.validate_arg(value)`

#### Cute Animal Picture

sadpuppy

### DIFF
--- a/wasm/datatypes/valtype.py
+++ b/wasm/datatypes/valtype.py
@@ -1,7 +1,19 @@
 import enum
+import sys
 
+from wasm import (
+    constants,
+)
+from wasm.exceptions import (
+    ValidationError,
+)
 from wasm.typing import (
+    Float32,
+    Float64,
+    TValue,
     UInt8,
+    UInt32,
+    UInt64,
 )
 
 from .bit_size import (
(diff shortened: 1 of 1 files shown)


wasm/datatypes/valtype.py
```diff
@@ -101,3 +113,34 @@ def get_integer_type(cls, num_bits: BitSize) -> 'ValType':
...
+    def validate_arg(self, arg):
+        if self is self.i32:
+            bounds = (0, constants.UINT32_MAX)
+            type_ = int
+        elif self is self.i64:
+            bounds = (0, constants.UINT64_MAX)
+            type_ = int
+        elif self in {self.f32, self.f64}:
+            # TODO: proper bounds
+            bounds = (sys.float_info.min, sys.float_info.max)
```
carver: If this isn't hard to track down in the spec, seems like worth adding now.

pipermerriam: I'm going to address it as my next task when I properly implement the floating point operations.  I tried to track down but it turns out that you have to pay money to get a copy of the IEEE-754 floating point standard: https://standards.ieee.org/standard/754-2008.html

PR #74, opened 2019-02-07
Convert parser typing to use typing.IO

pipermerriam:
## What was wrong?

All of the parsers were written using the `io.BytesIO` type when in fact we needed to use [`typing.IO[bytes]`](https://docs.python.org/3.5/library/typing.html#typing.io) to be able to support arbitrary file-like objects.

## How was it fixed?

Converted type hints for all parsers to use `IO[bytes]`

#### Cute Animal Picture

cone-of-shame-dog-vs-gloating-dog

### DIFF
--- a/wasm/parsers/blocks.py
+++ b/wasm/parsers/blocks.py
@@ -1,5 +1,5 @@
-import io
 from typing import (
+    IO,
     Tuple,
 )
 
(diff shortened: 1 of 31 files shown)


pipermerriam:
Merging without review since this is just a type checking cleanup.

PR #83, opened 2019-02-12
Finish rewriting legacy code

pipermerriam:
This builds on a few PRs that are open and ready for review.  Only the latest commit a7fcb43 is relevant.

### What is this

This finishes removing the legacy `wasm.main` module.

This includes 

- formalization of some mechanisms for instantiating wasm modules, and running module functions
- extracting all of the logic functions for individual opcodes
- expansion of the API around various datatypes to support the opcode logic functions.
- conversion of all value types to use `numpy` data types.
- proper implementation of the floating point operations.

3961c5b3470d771dc52da49915228183


tests/conftest.py
```diff
@@ -0,0 +1,3 @@
+def pytest_addoption(parser):
+    parser.addoption("--skip-slow-spec", action="store_true", required=False)
+    parser.addoption("--stop-after-command-line", type=int, required=False)
```
pipermerriam: This can be its own PR

wasm/datatypes/memory.py
```diff
@@ -22,4 +31,51 @@ class Memory(NamedTuple):
...
+        if end_location != int(location) + int(size):
+            raise Exception('err')
+        if end_location > len(self.data):
+            raise Trap(
+                f"Attempt to read from out of bounds memory location: {end_location} "
+                f"> {len(self.data)}"
+            )
+        return self.data[location:end_location]
+
+    def write(self, location: numpy.uint32, value: bytes) -> None:
```
pipermerriam: This functionality can be isolated into its own PR

wasm/instructions/numeric.py
```diff
@@ -460,11 +460,11 @@ class Extend(Interned):
     def __init__(self,
                  opcode: BinaryOpcode,
                  valtype: ValType,
-                 result: ValType,
+                 from_valtype: ValType,
                  signed: bool) -> None:
         self.opcode = opcode
         self.valtype = valtype
-        self.result = result
+        self.from_valtype = from_valtype
```
pipermerriam: This rename can get it's own PR

wasm/instructions/numeric.py
```diff
@@ -298,96 +298,96 @@ def __str__(self) -> str:
...
+            return cls(opcode, ValType.f64, None)
         elif opcode is BinaryOpcode.F64_MUL:
-            return cls(opcode, ValType.f64)
+            return cls(opcode, ValType.f64, None)
         elif opcode is BinaryOpcode.F64_DIV:
-            return cls(opcode, ValType.f64)
+            return cls(opcode, ValType.f64, None)
         elif opcode is BinaryOpcode.F64_MIN:
-            return cls(opcode, ValType.f64)
+            return cls(opcode, ValType.f64, None)
```
pipermerriam: This fix can get its own PR

wasm/logic/memory.py
```diff
@@ -0,0 +1,131 @@
...
+        else:
+            config.push_operand(instruction.valtype.value(raw_value))
+    elif instruction.valtype.is_float_type:
+        value = instruction.valtype.unpack_float_bytes(raw_bytes)
+        config.push_operand(value)
+    else:
+        raise Exception("Invariant")
+
+
+ENCODER_LOOKUP: Dict[BitSize, str] = {
```
pipermerriam: This can be removed.

wasm/logic/variable.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def get_global_op(config: Configuration) -> None:
+    instruction = cast(GlobalOp, config.instructions.current)
+    logger.debug("%s()", config.instructions.current.opcode.text)
+
+    global_address = config.frame.module.global_addrs[instruction.global_idx]
+    global_ = config.store.globals[global_address]
+    # TODO: remove runtime assertion
+    assert isinstance(global_.value, global_.valtype.value)
```
pipermerriam: remove

wasm/parsers/byte.py
```diff
@@ -1,28 +1,27 @@
...
     if byte:
-        return UInt8(byte[0])
+        return numpy.uint8(byte[0])
     else:
         raise ParseError("Unexpected end of stream")
 
 
 def parse_bytes(stream: IO[bytes]) -> bytes:
-    size = parse_u32(stream)
+    size = parse_size(stream)
```
pipermerriam: This can get its own PR

wasm/tools/fixtures/normalizers.py
```diff
@@ -66,28 +67,33 @@ def normalize_module_command(raw_command: RawCommand,
     )
 
 
+def _normalize_raw_value(valtype: ValType, raw_value: int) -> TValue:
+    if valtype is ValType.i32:
+        return numpy.uint32(raw_value)
+    elif valtype is ValType.i64:
+        return numpy.uint64(raw_value)
```
pipermerriam: condense to:

```python
if valtype.is_integer_type:
    return valtype.value(raw_value)
```

wasm/tools/fixtures/normalizers.py
```diff
@@ -66,28 +67,33 @@ def normalize_module_command(raw_command: RawCommand,
...
+    elif valtype is ValType.f32:
+        return valtype.unpack_float_bytes(numpy.uint32(raw_value).tobytes())
```
pipermerriam: condense to:

```python
elif valtype.is_float_type:
    return valtype.unpack_float_bytes(raw_value).tobytes()
```

wasm/datatypes/memory.py
```diff
@@ -22,4 +31,51 @@ class Memory(NamedTuple):
...
+            try:
+                end_location = location + size
+            except FloatingPointError:
+                raise Trap(
+                    f"Attempt to read from out of bounds memory location: {int(location) + size} "
+                    f"> {len(self.data)}"
+                )
+
+        if end_location != int(location) + int(size):
+            raise Exception('err')
```
pipermerriam: Coverage shows this code path isn't hit.  REMOVE

wasm/execution/configuration.py
```diff
@@ -85,6 +104,65 @@ def pop2_operands(self) -> Tuple[TValue, TValue]:
...
+
+    #
+    # Pop u64
+    #
+    def pop_u64(self) -> numpy.uint64:
+        return cast(numpy.uint64, self.frame.active_operand_stack.pop())
+
+    def pop2_u64(self) -> Tuple[numpy.uint64, numpy.uint64]:
+        a, b = self.frame.active_operand_stack.pop2()
+        return cast(numpy.uint64, a), cast(numpy.uint64, b)
```
pipermerriam: remove implementation, replace with `pass`

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+    a_s = numpy.int32(a)
+    logger.debug("%s(%s, %s)", config.instructions.current.opcode.text, a_s, b_s)
+
+    if a_s < b_s:
+        config.push_operand(constants.U32_ONE)
+    else:
+        config.push_operand(constants.U32_ZERO)
+
+
+def i32les_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version (and maybe unsigned version)

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+    if a >= b:
+        config.push_operand(constants.U32_ONE)
+    else:
+        config.push_operand(constants.U32_ZERO)
+
+
+#
+# 32-bit Signed integer comparisons
+#
+def i32lts_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version (and maybe unsigned version)

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+    a_s = numpy.int32(a)
+    logger.debug("%s(%s, %s)", config.instructions.current.opcode.text, a_s, b_s)
+
+    if a_s <= b_s:
+        config.push_operand(constants.U32_ONE)
+    else:
+        config.push_operand(constants.U32_ZERO)
+
+
+def i32gts_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version (and maybe unsigned version)

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+    a_s = numpy.int32(a)
+    logger.debug("%s(%s, %s)", config.instructions.current.opcode.text, a_s, b_s)
+
+    if a_s > b_s:
+        config.push_operand(constants.U32_ONE)
+    else:
+        config.push_operand(constants.U32_ZERO)
+
+
+def i32ges_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version (and maybe unsigned version)

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+    if a_s >= b_s:
+        config.push_operand(constants.U32_ONE)
+    else:
+        config.push_operand(constants.U32_ZERO)
+
+
+#
+# Integer addition
+#
+def i32add_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+
+def i64add_op(config: Configuration) -> None:
+    b, a = config.pop2_u64()
+    logger.debug("%s(%s, %s)", config.instructions.current.opcode.text, a, b)
+
+    with allow_overflow():
+        config.push_operand(a + b)
+
+
+def i32sub_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+    logger.debug("%s(%s, %s)", config.instructions.current.opcode.text, a, b)
+
+    with allow_overflow():
+        config.push_operand(a - b)
+
+
+#
+# Integer multiplication
+#
+def i32mul_op(config: Configuration) -> None:
```
pipermerriam: combine with 64 bit version

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+# Helpers for floating point ops
+#
+def _same_signed(left: TFloat, right: TFloat) -> bool:
+    left_is_negative = _is_negative(left)
+    right_is_negative = _is_negative(right)
+    return left_is_negative is right_is_negative
+
+
+def _same_signed_inf(left: TFloat, right: TFloat) -> bool:
+    if not numpy.isinf(left) or not numpy.isinf(right):
```
pipermerriam: update to use `numpy.isposinf`

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+
+
+def _different_signed(left: TFloat, right: TFloat) -> bool:
+    left_is_negative = _is_negative(left)
+    right_is_negative = _is_negative(right)
+    return left_is_negative is not right_is_negative
+
+
+def _different_signed_inf(left: TFloat, right: TFloat) -> bool:
+    if not numpy.isinf(left) or not numpy.isinf(right):
```
pipermerriam: update to use `numpy.isposinf`

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+#
+def fadd_op(config: Configuration) -> None:
+    instruction = cast(BinOp, config.instructions.current)
+
+    b, a = config.pop2_f64()
+
+    logger.debug("%s(%s, %s)", instruction.opcode.text, a, b)
+
+    with allow_overflow():
+        with allow_invalid():
```
pipermerriam: establish pattern for combining to reduce context overhead.

wasm/logic/numeric.py
```diff
@@ -0,0 +1,1071 @@
...
+
+
+def i32reinterpret_f32_op(config: Configuration) -> None:
+    instruction = cast(Convert, config.instructions.current)
+
+    value = config.pop_f32()
+
+    logger.debug("%s(%s)", instruction.opcode.text, value)
+
+    config.push_operand(numpy.frombuffer(value.tobytes(), numpy.uint32)[0])
```
pipermerriam: combine all reinterpret since we can do:

```python
numpy.frombuffer(value.tobytes(), instruction.valtype.value)[0]
```

wasm/logic/variable.py
```diff
@@ -0,0 +1,66 @@
...
+    GlobalOp,
+    LocalOp,
+)
+
+logger = logging.getLogger('wasm.logic.variable')
+
+
+def set_local_op(config: Configuration) -> None:
+    instruction = cast(LocalOp, config.instructions.current)
+    logger.debug("%s()", config.instructions.current.opcode.text)
```
pipermerriam: Update all opcode logging statements (where possible) to use static TEXT to reduce overhead.

wasm/datatypes/memory.py
pipermerriam: Except it couldn't because it was all wrapped up with the numpy stuff.

tests/conftest.py
pipermerriam: https://github.com/ethereum/py-wasm/pull/93

wasm/instructions/numeric.py
pipermerriam: https://github.com/ethereum/py-wasm/pull/89

wasm/parsers/byte.py
pipermerriam: https://github.com/ethereum/py-wasm/pull/90

wasm/logic/numeric.py
pipermerriam: Ended up not being quite that easy.

wasm/logic/numeric.py
pipermerriam: Ended up not being quite that simple

wasm/logic/variable.py
pipermerriam: Saving this for later cleanup performance improvements cleanup (trading tech debt for expediency)

stubs/numpy.py
```diff
@@ -0,0 +1,178 @@
...
+)
+
+from mypy_extensions import (
+    TypedDict,
+)
+
+T = TypeVar('T')
+
+
+class TErrorOptions(TypedDict):
```
davesque: Is there a reason to use a typed dict instead of a named tuple?  My initial reaction is that there's no real advantage since the performance characteristics of named tuples are inherently better and they seem to provide all the needed functionality.

stubs/numpy.py
```diff
@@ -0,0 +1,178 @@
...
+
+def trunc(v: T) -> T:
+    ...
+
+
+def round(v: T) -> T:
+    ...
+
+
+def sqrt(v: T) -> T:
```
davesque: We spoke about this in person, but just for the record this function is not always closed across a type for the possible values that might be passed to it.  This is probably also true for a couple of other functions in this module.  For example,
```python
def sqrt(n: np.int) -> np.float:
    return np.sqrt(n)
```
I guess for our purposes this might not matter, but just making a note of it.

tests/core/execution/test_configuration_object.py
```diff
@@ -0,0 +1,247 @@
...
+    config.pop_frame()
+
+    assert config.has_active_label is True
+
+    config.pop_label()
+
+    assert config.has_active_label is False
+
+
+def test_configuration_active_label_property(config):
```
davesque: Just wanted to mention that I appreciate the informative use of whitespace in this module.  I often notice a lack of that simple semantic tool in a lot of people's code.

tests/core/val-type/test_valtype_validate_argument.py
```diff
@@ -0,0 +1,55 @@
...
+)
+from wasm.datatypes import (
+    ValType,
+)
+from wasm.exceptions import (
+    ValidationError,
+)
+
+
+@pytest.mark.parametrize(
```
davesque: I've begun to think this is a bit of an anti-pattern.  If a parametrized test is testing two different classes of data (valid vs. invalid in this case), I think it's better to split them explicitly into two different parametrized tests instead of doing logic in the test function to distinguish them and requiring the inclusion of some marker value in the parameter tuple.

tests/spec/test_spec.py
```diff
@@ -16,13 +18,19 @@
 
 
 def pytest_generate_tests(metafunc):
```
davesque: Comment and not a complaint: this method of generating test parameters is interesting and I had no idea pytest had this API.  Cheers on this one!

tests/spec/test_spec.py
```diff
@@ -16,13 +18,19 @@
...
         metafunc=metafunc,
         fixtures_dir=FIXTURES_DIR,
+        skip_slow=skip_slow_spec,
     )
 
 
-def test_json_fixture(fixture_path):
+def test_json_fixture(fixture_path, pytestconfig):
+    # ensure that all numpy flags are set to raise exceptions
+    numpy.seterr(all='raise')
```
davesque: This is just a concern that may not be relevant.  I saw this line and decided to test some of the behavior myself just out of curiosity.  I was surprised to find out that numpy does *not* raise an exception when you supply a python value to one of its value constructors that would cause an overflow:
```
In [1]: import numpy as np

In [2]: np.seterr(all='raise')
Out[2]: {'divide': 'warn', 'over': 'warn', 'under': 'ignore', 'invalid': 'warn'}

In [3]: np.int8(128)
Out[3]: -128
```
Of course, it does raise an exception when an arithmetic operation causes an overflow:
```
In [4]: np.int8(64) * np.int8(2)
---------------------------------------------------------------------------
FloatingPointError                        Traceback (most recent call last)
<ipython-input-4-3712cff83f6c> in <module>
----> 1 np.int8(64) * np.int8(2)

FloatingPointError: overflow encountered in byte_scalars
```
Just thought I'd point out that behavior in case it was overlooked or in case you were acting on any incorrect assumptions.

wasm/constants.py
```diff
@@ -1,39 +1,53 @@
-from .typing import (
-    SInt32,
-    SInt64,
-    UInt32,
-    UInt64,
-)
+import numpy
 
-UINT6_CEIL = UInt32(2 ** 6)
+UINT6_CEIL = numpy.uint32(2 ** 6)
```
davesque: It might make sense to take advantage of numpy's built-in resources for getting these values.  For example,
```python
def get_int_range(dtype):
    iinfo = np.iinfo(dtype)
    return dtype(iinfo.min), dtype(iinfo.max)

def get_float_range(dtype):
    finfo = np.finfo(dtype)
    return finfo.min, finfo.max
```

davesque: One caveat with this is that it's getting the value ranges that are dependent upon the hardware.  But I think that's essentially the case anyway since you're storing the values with numpy and numpy just stores things at the hardware level.

wasm/datatypes/bit_size.py
```diff
@@ -1,8 +1,68 @@
...
-    b64 = 64
+    b8 = numpy.uint8(8)
+    b16 = numpy.uint8(16)
+    b32 = numpy.uint8(32)
+    b64 = numpy.uint8(64)
+
+    def unpack_int_bytes(self, raw_bytes: bytes, signed: bool) -> TUnpack:
+        if self is self.b8:
+            if signed:
+                return numpy.frombuffer(raw_bytes, numpy.int8)[0]
```
davesque: Might want to be careful about this since it's architecture dependent (big endian vs. little endian).

stubs/numpy.py
pipermerriam: This isn't our API but rather numpy's.  I used `TypedDict` so that we could have stronger safety on the actual return type.

https://docs.scipy.org/doc/numpy/reference/generated/numpy.seterr.html

stubs/numpy.py
pipermerriam: Yeah, I suspect we can tighten some of these up at some point but at this stage I just wanted to ensure that we didn't end up with what was happening previously, which was `mypy` thinking that the return types were plain `int` or `float`

tests/spec/test_spec.py
pipermerriam: This is the same pattern used in py-evm and seems to be appropriate and work well for situations like this where you want individual tests for some *unknown* amount of tests from an external source.

https://github.com/ethereum/py-evm/blob/100b8c7d5267a6a6c9ae9c0472963db217b390db/eth/tools/fixtures/generation.py#L43-L91

Also worth noting that it'd be possible to do this with a parametrized fixture which *might* be a little less magic but not by very much.

wasm/logic/control.py
```diff
@@ -0,0 +1,269 @@
...
+        _exit_block(config)
+    elif config.has_active_frame:
+        _return_from_function(config)
+    else:
+        raise Exception("Invariant?")
+
+
+def _br(config: Configuration, label_idx: LabelIdx) -> None:
+    label = config.get_by_label_idx(label_idx)
+    # take any return values off of the stack before poping labels
```
carver:
```suggestion
    # take any return values off of the stack before popping labels
```

davesque:
```suggestion
    # take any return values off of the stack before pooping labels
```

tests/core/execution/test_configuration_object.py
```diff
@@ -0,0 +1,247 @@
...
+    config.push_frame(frame_b)
+
+    assert config.has_active_label is False
+
+    # now pop the frame and should have an active label
+    config.pop_frame()
+
+    assert config.has_active_label is True
+
+    config.pop_label()
```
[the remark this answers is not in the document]
pipermerriam: I think the right thing to do is throw an error.  Updating API and adding a test.  good catch.

wasm/constants.py
pipermerriam: Yeah, I had originally done it something similar to this but I converted to hard coded values so that we couldn't end up in a situation where a hardware difference caused the library to malfunction.

pipermerriam:
I'll spend some time tomorrow filling in docstrings and will check in with you guys before merging.

PR #87, opened 2019-02-12
Memoize frame, label, operand stack, and instruction values

pipermerriam:
### What was wrong

Profiling showed that a **lot** of time was being spent 

- determining which operand stack and instruction sequence was active during execution.
- what the current instruction was
- what the current label was for a frame

This was due to these being implemented as property methods and them having to dig into other data structures.  Since these are accessed **alot** during execution, it seems to make sense to make some mutability trade-offs in this section of the code.

### How was it fixed

These values are now memoized at both the `Frame` and `Configuration` level as things are pushed and popped from the various stacks.

The following cleanups were also done:

- remove direct access to `Configuration` properties in most cases.
- remove un-used `Configuration` APIs
  - `config.active_label`
  - rename `get_by_label_idx` to `get_label_by_idx` because I have been consistently typing the later by instinct which indicates my initial naming was wrong.
- use a `memoryview` for reading memory sections to reduce number of object copies.
- change `MemoryInstance` to a normal object as `NamedTuple` no longer made sense with the following changes:
  - local cache of `len(data)`
- move `Exhaustion` error due to frame stack too large into `Configuration` API as it didn't feel appropriately placed within the control logic function code.
- fix error string when linking fails to use correct module name
- add ability to load directly from a file-like object on `Runtime` object (needed in ewasm)

fox-attacking-chic_1655097a

### DIFF
--- a/tests/core/execution/test_configuration_object.py
+++ b/tests/core/execution/test_configuration_object.py
@@ -56,7 +56,7 @@ def test_configuration_has_active_frame(config):
 
 
 def test_configuration_frame_property(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.frame
 
     frame_a = mk_frame()
@@ -121,13 +121,13 @@ def test_configuration_cannot_pop_frame_with_active_label(config):
 
 
 def test_configuration_active_label_property(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.active_label
 
     frame_a = mk_frame()
     config.push_frame(frame_a)
 
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.active_label
 
     label_a = mk_label()
@@ -144,7 +144,7 @@ def test_configuration_active_label_property(config):
     frame_b = mk_frame()
     config.push_frame(frame_b)
 
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.active_label
 
     label_c = mk_label()
@@ -154,7 +154,7 @@ def test_configuration_active_label_property(config):
 
 
 def test_configuration_instructions_property(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.instructions
 
     frame_a = mk_frame()
@@ -184,7 +184,7 @@ def test_configuration_instructions_property(config):
 
 
 def test_configuration_push_and_pop_from_operand_stack(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.push_operand(0)
 
     frame_a = mk_frame()
@@ -210,7 +210,7 @@ def test_configuration_push_and_pop_from_operand_stack(config):
 
 
 def test_configuration_push_and_pop_u32_from_operand_stack(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.push_operand(0)
 
     frame_a = mk_frame()
@@ -229,7 +229,7 @@ def test_configuration_push_and_pop_u32_from_operand_stack(config):
 
 
 def test_configuration_push_and_pop_u64_from_operand_stack(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.push_operand(0)
 
     frame_a = mk_frame()
@@ -248,7 +248,7 @@ def test_configuration_push_and_pop_u64_from_operand_stack(config):
 
 
 def test_configuration_get_label_by_idx(config):
-    with pytest.raises(IndexError):
+    with pytest.raises(AttributeError):
         config.push_operand(0)
 
     frame = mk_frame()
--- a/tests/core/execution/test_configuration_object.py
+++ b/tests/core/execution/test_configuration_object.py
@@ -24,22 +24,22 @@ def mk_label():
 
 
 def test_configuration_frame_stack_size(config):
-    assert config.frame_stack_size == 0
+    assert len(config._frame_stack) == 0
 
     frame_a = mk_frame()
     config.push_frame(frame_a)
 
-    assert config.frame_stack_size == 1
+    assert len(config._frame_stack) == 1
 
     frame_b = mk_frame()
     config.push_frame(frame_b)
 
-    assert config.frame_stack_size == 2
+    assert len(config._frame_stack) == 2
 
     assert config.pop_frame() is frame_b
     assert config.pop_frame() is frame_a
 
-    assert config.frame_stack_size == 0
+    assert len(config._frame_stack) == 0
 
 
 def test_configuration_has_active_frame(config):
@@ -56,22 +56,21 @@ def test_configuration_has_active_frame(config):
 
 
 def test_configuration_frame_property(config):
-    with pytest.raises(AttributeError):
-        config.frame
+    assert not hasattr(config, 'frame')
 
     frame_a = mk_frame()
     config.push_frame(frame_a)
 
-    assert config.frame is frame_a
+    assert config._frame is frame_a
 
     frame_b = mk_frame()
     config.push_frame(frame_b)
 
-    assert config.frame is frame_b
+    assert config._frame is frame_b
 
     assert config.pop_frame() is frame_b
 
-    assert config.frame is frame_a
+    assert config._frame is frame_a
 
 
 def test_configuration_has_active_label(config):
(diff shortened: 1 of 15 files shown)


pipermerriam:
Noting that we could also do this with the *current* instruction to reduce lookups on `config.instructions.current`

pipermerriam:
Master performance:

```
thunk()                x      10000: 0.6985  |  0.0001 per-call
add(1, 2)              x      10000: 1.1606  |  0.0001 per-call
call_thunk(10000,)     x          1: 1.2929  |  1.2929 per-call
call_add(10000, 1, 2)  x          1: 1.9410  |  1.9410 per-call
wrapped thunk: 0.00012929 per call
wrapped add  : 0.00019410 per call
call overhead for thunk: 0.5943 (85.09%)  |   0.00005943 per call
call overhead for add  : 0.7804 (67.25%)  |   0.00007804 per call
```

This branch

```
thunk()                x      10000: 0.5913  |  0.0001 per-call
add(1, 2)              x      10000: 0.9960  |  0.0001 per-call
call_thunk(10000,)     x          1: 0.9949  |  0.9949 per-call
call_add(10000, 1, 2)  x          1: 1.4550  |  1.4550 per-call
wrapped thunk: 0.00009949 per call
wrapped add  : 0.00014550 per call
call overhead for thunk: 0.4036 (68.25%)  |   0.00004036 per call
call overhead for add  : 0.4589 (46.08%)  |   0.00004589 per call
```

carver:
Nice little 2x performance bump! :)

wasm/execution/configuration.py
```diff
@@ -218,8 +214,11 @@ def __init__(self, store: Store) -> None:
     def execute(self) -> Tuple[TValue, ...]:
         from wasm.logic import OPCODE_TO_LOGIC_FN
 
-        while self.has_active_frame:
-            instruction = next(self.instructions)
+        while True:
+            try:
+                instruction = next(self.instructions)
+            except AttributeError:
```
carver: Seems a little weird in this context to catch an `AttributeError`. It requires knowing the internals of how the iterator works. Maybe catch and re-raise a different exception inside the iterator? We could potentially even just raise a `StopIteration`, then this whole loop reduces to: `instruction = last(self.instructions)`

wasm/execution/configuration.py
```diff
@@ -218,8 +214,11 @@ def __init__(self, store: Store) -> None:
...
+                break
 
             logic_fn = OPCODE_TO_LOGIC_FN[instruction.opcode]
```
carver: Looks like if there is an `AttributeError` right away, then `instruction` hasn't been assigned yet. Maybe that's fine if we have proven that impossible during a validation step? (although probably worth a comment, then)

wasm/execution/instructions.py
```diff
@@ -18,10 +18,15 @@ class InstructionSequence(Sequence):
...
     """
     _instructions: Tuple[BaseInstruction, ...]
+    current: BaseInstruction
 
     def __init__(self, instructions: Iterable[BaseInstruction]) -> None:
         self._instructions = tuple(instructions)
         self._idx = -1
+        try:
+            self.current = self._instructions[0]
+        except IndexError:
```
carver: If instructions is empty in some normal cases, it seems like it must be faster to do:

```py
if self._instructions:
  self.current = self._instructions[0]
```

If it is never supposed to happen, then why squash the exception?

wasm/execution/configuration.py
pipermerriam: I've updated a number of things, but kept this section the same for the following reasons.

The `AttributeError` now actually comes from access to `self._instructions`.  I've moved the `next` (now `__next__()`) call down outside of the `try/except` to make it clear where the exception comes from.

The reason for this is that I've now extended the memoization to the `Configuration` object itself to reap some further performance gains.  Part of this was changing from use of an `@property` method to a direct property to reduce the overhead of the additional call frame caused by computed properties.

I've added a large comment in the body of this function outlining the parts that are less readable or *pythonic* in favor of performance to make it clearer why it is written the way that it is.

wasm/execution/configuration.py
pipermerriam: Maybe I'm mis-understanding here, but the `try/except` above breaks out of the loop on attribute error which should give us the strong guarantee that access to the local `instruction` variable can only happen in the case that the next instruction was pulled from the instruction sequence.

wasm/execution/instructions.py
pipermerriam: I've actually removed the memoization of current instruction from the sequence object and moved it down into the `Configuration` object which removes the need to check for this edge case.

pipermerriam:
@carver in the process of addressing your feedback I've made substantive changes that warrant another full pass.

wasm/execution/configuration.py
carver: Hmmm, must have been a brain-fart. I am guessing I thought it was on a different indentation. :+1:

wasm/execution/instructions.py
pipermerriam: stock-vector-cartoon-brain-having-a-brain-fart-vector-clip-art-illustration-with-simple-gradients-all-in-a-151038323

wasm/datatypes/memory.py
```diff
@@ -47,21 +53,22 @@ def read(self, location: numpy.uint32, size: numpy.uint32) -> bytes:
...
                 )
 
-        if end_location > len(self.data):
+        if end_location > self._length_cache:
             raise Trap(
                 f"Attempt to read from out of bounds memory location: {end_location} "
                 f"> {len(self.data)}"
             )
-        return self.data[location:end_location]
+        return memoryview(self.data)[location:end_location]
```
carver: Hm, it makes me a little queasy that this enables readers to modify the `MemoryInstance`s `.data` with this returned `memoryview`. I tried to find a clean way to make the `memoryview` readonly, but got stuck. [Another person got stuck in the same way](https://lists.gt.net/python/python/1124122#1124122) and resorted to this mess:

```py

pydll = ctypes.cdll.LoadLibrary("python{}{}".format( 
  sys.version_info.major, sys.version_info.minor
)) 

def ro_memoryview_from_bytearray(buffer): 
  assert isinstance(buffer, bytearray) 
  ptr = ctypes.c_char_p(pydll.PyByteArray_AsString( 
    ctypes.py_object(buffer)
  )) 
  mv_id = pydll.PyMemoryView_FromMemory(ptr, len(buffer), 0) 
  return ctypes.cast(mv_id, py_object).value

File: wasm/execution/configuration.py [hunk 5]
```diff
@@ -268,106 +304,121 @@ def extend_results(self, values: Iterable[TValue]) -> None:
...
     #
-    @property
-    def frame_stack_size(self) -> int:
-        return len(self.frame_stack)
-
     def push_frame(self, frame: Frame) -> None:
-        self.frame_stack.push(frame)
+        if len(self._frame_stack) > 1024:
+            # This is not part of spec, but this is required to pass tests.
+            # Tests pass with limit 10000, maybe more
```
carver: Didn't fully grok this comment. The tests would pass if you check `len(self._frame_stack) > 10000`? Then where did `1024` come from?

Is there a place where we can report that the tests don't seem to match the spec?

wasm/execution/configuration.py
```diff
@@ -268,106 +304,121 @@ def extend_results(self, values: Iterable[TValue]) -> None:
...
+            raise Exhaustion("Too many call frames.  Cannot exceed 1024")
+
+        self._frame = frame
+        self._frame_stack.push(frame)
+        self._operand_stack = self._frame.active_operand_stack
+        self._instructions = self._frame.active_instructions
```
carver: If this is a super-performance-sensitive area, then maybe skip the extra attribute lookup and do:
```suggestion
        self._instructions = frame.active_instructions
```

tests/core/execution/test_configuration_object.py
```diff
@@ -120,71 +119,37 @@ def test_configuration_cannot_pop_frame_with_active_label(config):
         config.pop_frame()
 
 
-def test_configuration_active_label_property(config):
```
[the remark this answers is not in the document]
pipermerriam: This is the API on the `Configuration` object (not the `Frame` like I think you're thinking).  I realized nothing was using this API so I dumped it.

wasm/datatypes/memory.py
pipermerriam: https://github.com/ethereum/py-wasm/issues/115

wasm/execution/configuration.py
pipermerriam: https://github.com/ethereum/py-wasm/issues/116

PR #90, opened 2019-02-12
Add parse_size utility

pipermerriam:
## What was wrong?

Once the `numpy` stuff lands, `mypy` complains about using the numpy integer types in `stream.read()`.

## How was it fixed?

Added a `parse_size` parser that returns the value as a plain `int`

#### Cute Animal Picture

hqdefault

### DIFF
--- a/wasm/parsers/size.py
+++ b/wasm/parsers/size.py
@@ -1,4 +1,7 @@
-from typing import IO
+from typing import (
+    IO,
+    cast,
+)
 
 from .integers import (
     parse_u32,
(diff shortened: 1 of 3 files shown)


wasm/parsers/size.py
```diff
@@ -0,0 +1,10 @@
+from typing import IO
+
+from .integers import (
+    parse_u32,
+)
+
+
+def parse_size(stream: IO[bytes]) -> int:
```
davesque: Is this to satisfy the type checker?  If so, would it be more appropriate to use `cast` instead of creating a whole new `int` object from an existing one?

pipermerriam: I believe I may have actually hit runtime errors when I fed a `numpy` integer type to `stream.read` but I'll double check.  If there are no runtime errors, I'll convert to `cast`

pipermerriam: No runtime errors, converted to a `cast`.

PR #92, opened 2019-02-12
Cleanup unused APIs on Export type

pipermerriam:
## What was wrong?

There were multiple property methods on the `Export` type which are not used (as measured by running code coverage)

## How was it fixed?

Removed them since they weren't needed.

#### Cute Animal Picture

bird-w-flower-on-head


pipermerriam:
If there's anything you'd like cleaned up I'll do it post merge.

PR #93, opened 2019-02-12
Add new cli flags for easier spec test running

pipermerriam:
## What was wrong?

When debugging I got frustrated with 1) inability to skip the slowest tests and 2) having to inject custom code to stop the test run after a specific command line.

## How was it fixed?

Added two new CLI flags that can be used during test runs

- `--stop-after-command-line=<number>`
- `--skip-slow-spec`

#### Cute Animal Picture

4xrbfsw

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,3 @@
+def pytest_addoption(parser):
+    parser.addoption("--skip-slow-spec", action="store_true", required=False)
+    parser.addoption("--stop-after-command-line", type=int, required=False)
--- a/README.md
+++ b/README.md
@@ -96,3 +96,39 @@ If you are in a beta version, `make release bump=stage` will switch to a stable.
 
 To issue an unstable version when the current version is stable, specify the
 new version explicitly, like `make release bump="--new-version 4.0.0-alpha.1 devnum"`
+
+
+## Development and Testing
+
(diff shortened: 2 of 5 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,3 @@
+def pytest_addoption(parser):
+    parser.addoption("--skip-slow-spec", action="store_true", required=False)
+    parser.addoption("--stop-after-command-line", type=int, required=False)
```
carver: So this one is primarily meant to be used when you're running a single test, and you are only interested in the first part of that test running? Seems pretty specific, but I guess I'm missing a common case. A bit more info about how/when to use it would help out people who haven't been living/breathing the test suite. :)

pipermerriam: I'll add some developer docs about running the spec tests before merging with examples of how to use these commands.

pipermerriam: As for *specificity*, a common case is that a spec test is failing in the `i32.wast` tests on command line 123 in a branch you are working on (but passing in master).  You run `pytest tests/spec/ -k "132.wast" --stop-after-command-line=123 --log-cli-level=debug` in `master`, save the log output, do the same in your branch, and then compare the results.

Without this, the log statements for command line 123 will be *very* deep in log history as many of these files have 1,000's of commands/lines.

PR #99, opened 2019-02-14
Enable the assert NAN tests and get them passing

### LINKED ISSUE #94: Enable assertarithmaticnan spec tests
https://github.com/ethereum/py-wasm/blob/6559ef319126b4874a03ba9468e9d1eb7e066022/wasm/tools/fixtures/runner.py#L305-L318

Just realized there's a set of spec tests that are still not implemented.

pipermerriam:
fixes #94 

## What was wrong?

There were still some spec tests being skiped, specifically the ones that do assertions about returned NAN values.

## How was it fixed?

Enabled the tests and fixed the logic function implementations so that they passed.

#### Cute Animal Picture

animal-animals-babies-bird-birds-cute-favim com-40474

### DIFF
--- a/wasm/_utils/float.py
+++ b/wasm/_utils/float.py
@@ -1,0 +1,116 @@
+from typing import (
+    Any,
+    Tuple,
+)
+
+import numpy
+
+from wasm import (
+    constants,
+)
+from wasm.typing import (
+    AnyFloat,
+    AnyUnsigned,
+)
+
+# the mantissa bits
+MANTISSA_64_MASK = numpy.uint64(2**constants.F64_SIGNIF - 1)
+MANTISSA_32_MASK = numpy.uint32(2**constants.F32_SIGNIF - 1)
+
+# the expon bits
+EXPONENT_64_SHIFT = 64 - 1 - constants.F64_EXPON
+EXPONENT_32_SHIFT = 32 - 1 - constants.F32_EXPON
+EXPONENT_64_MASK = numpy.uint64(2**constants.F64_EXPON - 1 << EXPONENT_64_SHIFT)
+EXPONENT_32_MASK = numpy.uint32(2**constants.F32_EXPON - 1 << EXPONENT_32_SHIFT)
+
+# most significant bit
+SIGN_64_MASK = numpy.uint64(2**63)
+SIGN_32_MASK = numpy.uint32(2**31)
+
+_CHECK_64_AND_MASK = MANTISSA_64_MASK & EXPONENT_64_MASK & SIGN_64_MASK
+_CHECK_32_AND_MASK = MANTISSA_32_MASK & EXPONENT_32_MASK & SIGN_32_MASK
+_CHECK_64_OR_MASK = MANTISSA_64_MASK | EXPONENT_64_MASK | SIGN_64_MASK
+_CHECK_32_OR_MASK = MANTISSA_32_MASK | EXPONENT_32_MASK | SIGN_32_MASK
+
+
+assert _CHECK_64_AND_MASK == 0
+assert _CHECK_32_AND_MASK == 0
+assert _CHECK_64_OR_MASK == numpy.uint64(2**64 - 1)
+assert _CHECK_32_OR_MASK == numpy.uint32(2**32 - 1)
+
+
+def _decompose_float32(value: numpy.float32) -> Tuple[numpy.uint32, numpy.uint32, numpy.uint32]:
+    as_uint32 = numpy.frombuffer(value.data, numpy.uint32)[0]
+    sign = numpy.uint32(bool(as_uint32 & SIGN_32_MASK))
+    exponent = numpy.uint32(int(as_uint32 & EXPONENT_32_MASK) >> EXPONENT_32_SHIFT)
+    mantissa = as_uint32 & MANTISSA_32_MASK
+
+    return (sign, exponent, mantissa)
+
+
+def _decompose_float64(value: numpy.float64) -> Tuple[numpy.uint64, numpy.uint64, numpy.uint64]:
+    as_uint64 = numpy.frombuffer(value.data, numpy.uint64)[0]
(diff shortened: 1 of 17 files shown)


wasm/_utils/float.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+import numpy
+
+from wasm import (
+    constants,
+)
+from wasm.typing import (
+    AnyFloat,
+    AnyUnsigned,
```
carver: Maybe `UnsignedFloat`? `Any` name still makes me think it includes integers.

wasm/_utils/float.py
```diff
@@ -0,0 +1,116 @@
...
+        (int(sign) << 63) | (int(exponent) << EXPONENT_64_SHIFT) | int(mantissa)
+    )
+    return numpy.frombuffer(as_uint64.data, numpy.float64)[0]
+
+
+def compose_float32(sign: AnyUnsigned,
+                    exponent: AnyUnsigned,
+                    mantissa: AnyUnsigned) -> numpy.float32:
+    as_uint32 = numpy.uint32(
+        (int(sign) << 32) | (int(exponent) << EXPONENT_32_SHIFT) | int(mantissa)
```
carver: Should this be a 31 bit shift? If so, maybe there's a test case missing.

wasm/_utils/float.py
```diff
@@ -0,0 +1,116 @@
...
+        return _decompose_float32(value)
+    else:
+        raise TypeError(f"Invalid type: {type(value)}")
+
+
+def compose_float64(sign: AnyUnsigned,
+                    exponent: AnyUnsigned,
+                    mantissa: AnyUnsigned) -> numpy.float64:
+    as_uint64 = numpy.uint64(
+        (int(sign) << 63) | (int(exponent) << EXPONENT_64_SHIFT) | int(mantissa)
```
carver: Would be nice to move the sign shift magic number into a constant, both here and in float32.

wasm/logic/numeric.py
```diff
@@ -710,7 +710,8 @@ def fceil_op(config: Configuration) -> None:
     logger.debug("%s(%s)", instruction.opcode.text, value)
 
     if numpy.isnan(value):
-        config.push_operand(value)
+        with allow_invalid():
+            config.push_operand(numpy.ceil(value))
```
carver: Huh, interesting, does `ceil` actually do something here? Or is this equivalent to `push_operand(F64_CANON_NAN)`?

Just not sure what allowing invalid means in this context. If it's a no-op then it seems like it's not worth even running the `numpy.ceil()` function.

wasm/logic/numeric.py
```diff
@@ -1137,7 +1152,15 @@ def f64promote_op(config: Configuration) -> None:
 
     logger.debug("%s(%s)", instruction.opcode.text, value)
 
-    config.push_operand(numpy.float64(value))
+    if numpy.isnan(value):
+        if _is_negative(value):
+            logger.info('NEGATIVE')
+            config.push_operand(numpy.float64('-nan'))
+        else:
+            logger.info('POSITIVE')
```
carver: Looks like some cruft to be removed.

wasm/tools/fixtures/runner.py
```diff
@@ -172,14 +176,18 @@ def run_opcode_action_get(action: Action,
...
+
+
+def do_assert_return(command: TDoAssertReturnCommands,
                      module: CurrentModule,
                      all_modules: AllModules,
                      runtime: Runtime) -> None:
-    try:
-        ret = run_opcode_action(command, module, all_modules, runtime)
-    except FloatingPointNotImplemented:
-        return
```
carver: :tada:

wasm/_utils/float.py
davesque: Yeah, it should be.

wasm/_utils/float.py
```diff
@@ -0,0 +1,116 @@
...
+    AnyFloat,
+    AnyUnsigned,
+)
+
+# the mantissa bits
+MANTISSA_64_MASK = numpy.uint64(2**constants.F64_SIGNIF - 1)
+MANTISSA_32_MASK = numpy.uint32(2**constants.F32_SIGNIF - 1)
+
+# the expon bits
+EXPONENT_64_SHIFT = 64 - 1 - constants.F64_EXPON
```
davesque: Alternatively, you could just use `constants.FXX_SIGNIF` for these values.

wasm/_utils/float.py
```diff
@@ -0,0 +1,116 @@
...
+    else:
+        raise TypeError(f"Invalid type: {type(value)}")
+
+
+# runtime check since the actual value *might* differ on different platforms.
+assert is_canonical_nan(constants.F32_CANON_NAN)
+assert is_canonical_nan(constants.F64_CANON_NAN)
+
+
+def is_arithmetic_nan(value: Any) -> bool:
```
davesque: How is this function used?  It appears to leave out a range of nan values whose mantissas are less than `FXX_CANON_N` but greater than zero.

wasm/_utils/float.py
carver: Maybe a quick hypothesis test to make sure that `compose(decompose(afloat)) == afloat`

wasm/_utils/float.py
pipermerriam: It's from this section:

https://webassembly.github.io/spec/core/bikeshed/index.html#floating-point%E2%91%A0

> An arithmetic `NaN` is a floating-point value `±nan(n)` with `n≥canonN`, such that the most significant bit is `1` while all others are arbitrary.

wasm/logic/numeric.py
pipermerriam: The `invalid` comes from feeding a `NaN` into `numpy.ceil()`.  Internally, I don't know what the call does, but the tests fail without it.

PR #102, opened 2019-02-15
Use dtype data property over tobytes method when possible

### LINKED ISSUE #100: For frombuffer calls use `dtype.data` instead of `dtype.tobytes()`
The `numpy.frombuffer` API supports passing in `dtype.data` which is a `memoryview` which should remove some object allocation overhead.  Everywhere that we currently use `numpy.frombuffer(value.tobytes(), ...)` should be updated to use `numpy.frombuffer(value.data, ...)`

pipermerriam:
fixes #100 

builds on #99

## What was wrong?

We were incurring an extra object allocation by using `dtype.tobytes()` in places where we could have been directly using the `dtype.data` property.

## How was it fixed?

Converted to use the `.data` property when possible.

#### Cute Animal Picture

3ec499603ef8c7e43ff050bb302f543e

### DIFF
--- a/wasm/datatypes/valtype.py
+++ b/wasm/datatypes/valtype.py
@@ -260,7 +260,7 @@ class ValType(enum.Enum):
         else:
             raise TypeError(f"Cannot convert {self} to unsigned integer")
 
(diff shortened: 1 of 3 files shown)


pipermerriam:
Only 0550fda is relevant.

PR #104, opened 2019-02-15
[WIP] S-expression parsing of text modules

pipermerriam:
## What was wrong?

Issue #

## How was it fixed?

Summary of approach.

#### Cute Animal Picture


wasm/sexpressions.py
```diff
@@ -0,0 +1,232 @@
...
+exkind:  ( func <var> )
+         ( global <var> )
+         ( table <var> )
+         ( memory <var> )
+
+module:  ( module <name>? <typedef>* <func>* <import>* <export>* <table>? <memory>? <global>* <elem>* <data>* <start>? )
+         <typedef>* <func>* <import>* <export>* <table>? <memory>? <global>* <elem>* <data>* <start>?  ;; =
+         ( module <typedef>* <func>* <import>* <export>* <table>? <memory>? <global>* <elem>* <data>* <start>? )
+"""
+from pyparsing import (
```
davesque: We're already using [parsimonious](https://github.com/erikrose/parsimonious) in eth-abi ([here](https://github.com/ethereum/eth-abi/blob/master/eth_abi/grammar.py#L14-L38)) which is a little more user friendly IMHO.  Any thoughts about using that?  The nice thing about parsimonious is that you can define grammars using BNF-like syntax.

It's also pretty easy to convert parsing results into something usable via a `parsimonious.NodeVisitor` class.

pipermerriam: `pyparsing` was just what turned up when I was digging for reference implementations of how to parse S-Expressions.  :+1: to `parsimonious` if you like it better.

davesque:
Was just taking a look at this.  It looks like you've written a grammar which is pretty explicit.  Is there a reason you didn't go with a more generic s-expression grammar and then verify s-expression content as a kind of post-processing step?  I'm wondering if that wouldn't be simpler and allow you to forgo any possible headaches relating to writing a lengthy grammar.

davesque:
Actually, reviewing the WASM spec now, it looks like the text grammar is pretty well-defined.  So I guess that's probably one reason you did it that way.

pipermerriam:
> Actually, reviewing the WASM spec now, it looks like the text grammar is pretty well-defined. So I guess that's probably one reason you did it that way.

Yeah, I took their provided grammar and am converting it bit-by-bit to parsimonious.

I had that thought yesterday (use a generic s-expression grammar), but I think I like the up-front validation that this explicit grammar provides.

PR #107, opened 2019-02-26
Piper/extend benchmark suite

pipermerriam:
Builds on #106 

## What was wrong?

#106 introduced a benchmark suite *however* it was very limited, only containing two tests.  `add(1, 2)` and `thunk()`.

Ideally we want something more comprehensive that covers operations that properly emulate real world execution.

## How was it fixed?

The EWASM team pointed me at https://github.com/fluencelabs/fluence/tree/master/bench/vm

I ported over a few of their benchmarks.

- recursive `keccak` hashing for depths of 1, 10, 100
- recursive fibonacci calculation of the series numbers 1, 11, 17, 38
- factorization of prime numbers 1117, 9973, 2147483647

Each of these is actually a rust implementation compiled to wasm.

#### Cute Animal Picture

fox-and-chicken_grande


scripts/benchmark/run.py
```diff
@@ -0,0 +1,275 @@
...
+    for export in module.exports:
+        if export.is_function and export.name == function_name:
+            return export.function_address
+    else:
+        raise Exception(f"No function found by name: {function_name}")
+
+
+def do_basic_benchmarks():
+    runtime = Runtime()
+    module, _ = runtime.instantiate_module(runtime.load_module(BASIC_WASM))
```
carver: Worth benchmarking load time?

scripts/benchmark/run.py
```diff
@@ -0,0 +1,275 @@
...
+    if order == 1:
+        module, _ = runtime.instantiate_module(runtime.load_module(FIBONACCI_1_WASM))
+    elif order == 11:
+        module, _ = runtime.instantiate_module(runtime.load_module(FIBONACCI_11_WASM))
+    elif order == 17:
+        module, _ = runtime.instantiate_module(runtime.load_module(FIBONACCI_17_WASM))
+    elif order == 38:
+        module, _ = runtime.instantiate_module(runtime.load_module(FIBONACCI_38_WASM))
+    else:
+        raise Exception("Only 1, 11, 17, 38 are supported")
```
carver: I'm 98% on board with how much explicit validation we do, but every once in a while I think we might go a bit overboard. This is one of those where I think: "how much worse would it be to dynamically do this and raise on failure?" like:

```py
def load_fib(order):
  fib_file = BASE_DIR / 'fibonacci_{order}_bigint_wasm.wasm'
  if not fib_file.exists():
    raise Exception("fibonacci {order} is not supported: {fib_file} not found")
  module = runtime.load_module(fib_file)
  return runtime.instantiate_module(module)
```
Then `do_fibonacci_bigint_benchmarks` gets simpler. Anyway, just a nitpick that pops in my head occasionally.

scripts/benchmark/run.py
pipermerriam: Yes, will do in another PR

tracking issue: https://github.com/ethereum/py-wasm/issues/113

scripts/benchmark/run.py
pipermerriam: :+1: at this point a lot of this is reflex/habit/well-worn-ruts for me.  Always happy for pushback as I get blind to this sometimes.  will simplify.

pipermerriam:
Yes, you did review one, but it only benchmarked a `thunk` and single `add(1, 2)` operation.  I added these once I started into the deeper optimizations and realized the existing benchmarks were not useful for much beyond benchmarking *meta* stuff, but that they'd never really show performance of the actual execution.

PR #110, opened 2019-02-26
[WIP] use is checks over isinstance checks

pipermerriam:
## What was wrong?

`isinstance` is more expensive than doing a direct comparison of whether `type(obj) is SomeClass` as long as there is no need to deal with subclasses.  Most of the places in the codebase that do an `isinstance` don't need to deal with subclasses.

## How was it fixed?

Converted *most* of the places that use `isinstance` to instead use a direct `is` comparison.

#### Cute Animal Picture


pipermerriam:
Hrm, this has the unfortunate side effect of making `mypy` very angry.

https://circleci.com/gh/ethereum/py-wasm/2401?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Given that it only produced a minor performance gain since most of the checks occur during module loading and not execution, closing and tabling this approach for now.

